Issue #4, opened 2018-07-05
make sure that HLX env variables act as defaults


trieloff:
fix and test that
- `HLX_AUTH`
- `HLX_APIHOST` ?
- `HLX_LOGKEY`
- `HLX_LOGHOST` ?

act as defaults for the respective params.

@trieloff why don't we name it `HLX_WSKAUTH` and `HLX_WSKAPIHOST` and `HLX_WSKNAMESPACE` and/or support for `.wskprops` ?

----

We will have keys/auth for at least:

- OpenWhisk
- Loggly
- Fastly

So having it consistent as:

- `HLX_WSK_AUTH` + `HLX_WSK_HOST`, `HLX_WSK_NAMESPACE`
- `HLX_LOGGLY_AUTH` + `HLX_LOGGLY_HOST`
- `HLX_FASTLY_AUTH` + `HLX_FASTLY_HOST`

might be the best option.

PR #20, opened 2018-07-08
Cleanup deploy parameters

trieloff:
As discussed in #4


test/testDeployCmd.js
```diff
@@ -0,0 +1,30 @@
...
+    new CLI()
+      .run(['deploy',
+        '--wsk-auth', 'secret-key',
+        '--wsk-namespace', 'hlx',
+        '--no-auto',
+        '--dirty',
+        '--dry-run',
+        '--target', 'test/integration/.hlx/build',
+      ]);
+    done();
```
tripodsan: done should be invoked in the `then` of the run. and `.catch(done)` is missing.

test/testDeployCli.js
```diff
@@ -68,42 +69,65 @@ describe('hlx deploy', () => {
...
+    sinon.assert.calledWith(mockDeploy.withLogglyAuth, '');
+    sinon.assert.calledWith(mockDeploy.withTarget, '.hlx/build');
+    sinon.assert.calledWith(mockDeploy.withDocker, 'trieloff/custom-ow-nodejs8:latest');
+    sinon.assert.calledWith(mockDeploy.withPrefix, 'git-github-com-example-project-helix--master--');
+    sinon.assert.calledWith(mockDeploy.withDefault, undefined);
+    sinon.assert.calledOnce(mockDeploy.run);
+  });
+
+  it('hlx deploy works with arguments provided in environment', () => {
+    process.env.HLX_WSK_AUTH = 'sekret-key';
```
tripodsan: test all other possible arguments als well

test/testDeployCmd.js
trieloff: `run()` does not return a promise, it returns the parsed `args`.

test/testDeployCli.js
trieloff: I think that would be overkill, as the `env()` method of `yargs` is well-tested and we won't get any more functional coverage: https://github.com/yargs/yargs/blob/6ee2c82df515cb1b062e4135a5dd9c386fed2b21/test/yargs.js#L756

PR #44, opened 2018-07-19
hlx strains creates entire Service Config now

trieloff:
As @kptdobe reported, setting up a new Fastly configuration from scratch is a task that is better left automated, so here is a first shot at automating it:

- defines backends for GitHub and Adobe I/O Runtime
- sets `helix.vcl` (what used to be `primordialsoup.vcl`) – can be overridden
- works better when VCL files have not been created before
- `run()` waits for the configuration to complete before returning
- comes with tests for updating a new service config and an existing service config


test/mocha.opts
```diff
@@ -0,0 +1 @@
+--timeout 15000
```
tripodsan: why is this here? is this global for all tests? it's better to define it for individual tests or for suites:
see https://mochajs.org/#timeouts

src/strain.cmd.js
```diff
@@ -21,6 +21,8 @@ const { toBase64 } = require('request/lib/helpers');
 const strainconfig = require('./strain-config-utils');
 
 const STRAIN_FILE = path.resolve(process.cwd(), '.hlx', 'strains.yaml');
+const HELIX_VCL_FILE = path.resolve(process.cwd(), 'helix.vcl');
+const HELIX_VCL_DEFAULT_FILE = path.resolve(__dirname, 'helix.vcl');
```
tripodsan: can we move this outside the normal `src` folder? maybe to `layouts` ?

src/helix.vcl
```diff
@@ -0,0 +1,1216 @@
...
+          <div class='_1'>SERVER ERROR</div>
+          <code class='btn'>There is no action with the configured prefix.</code>
+      </div>
+      </body>
+    </html>"};
+    return(deliver);
+  }
+  if (obj.status == 953 ) {
+    set obj.http.Content-Type = "text/html";
+    synthetic {"<!doctype html>
```
tripodsan: is there a better way to do this? maybe use some sort of `include` ?
having this large vlc is rather hard to understand and maintain. also, inline HTML is rather nasty.

src/strain.cmd.js
```diff
@@ -278,29 +356,53 @@ class StrainCommand {
...
+      // overwrite
+      Object.assign({
+        method: 'PUT',
+        form: {
+          name,
+          content: vcl,
+        },
+      }, baseopts)).catch(() =>
+      // create new
+      Object.assign({
```
tripodsan: you should probably test whit the request failed, and only create a new one, for a 404.

src/strain.cmd.js
```diff
@@ -343,86 +461,87 @@ class StrainCommand {
...
-    return this;
+      // wait for the two initialization jobs to finish, too
+      strainjobs.push(initfp);
+      strainjobs.push(initdp);
+
+      // also wait for the strain.vcl writing to finish
+      strainjobs.push(strainp);
+      strainjobs.push(secretp);
+
+      return Promise.all(strainjobs).then(() => {
```
tripodsan: btw `Promise.all()` does not process the jobs sequentially... don't know if this is important.

trieloff: I know, and I'm fine with that.

test/mocha.opts
trieloff: Yes, this sets the global timeout. I got impatient after chasing timeout issue after timeout issue.

src/strain.cmd.js
trieloff: Will do.

src/helix.vcl
trieloff: I think I can generate the `synthetic` block from example html files and then include it. Only downside is that we will have to make additional requests to create these VCL files.

src/strain.cmd.js
trieloff: Will do.

trieloff:
Let's not move the functions out into a util class until we use them in more than one place.

PR #47, opened 2018-07-20
Faster build

### LINKED ISSUE #40: Run tests in parallel
As we get more tests, let's see if we can run them in parallel to speed up CI

trieloff:
Fixes #40 

- use 10 workers instead of 1
- test and get code coverage in one step
- publish examples in a different workflow step

brings down build time from >2min to <1min


test/testStrainCmd.js
```diff
@@ -22,10 +23,13 @@ const WSK_AUTH = 'nope';
 
 describe('hlx strain (Integration)', () => {
   beforeEach(() => {
+    fs.mkdirpSync('.hlx');
```
tripodsan: I would not do this in the project directory. use the `test/tmp` or something

trieloff:
I won't get this done before I leave for vacation, so feel free to make changes.

Issue #100, opened 2018-08-17
[hlx up] exits quietly in a non-git directory


stefan-guggisberg:
e.g.

```bash
mkdir test && cd test
hlx up
```
doesn't do anything. it should show an error instead.

PR #101, opened 2018-08-17
Fix and test for #100

trieloff:
See #100: it looks like the constructor can fail, too. I've used the opportunity to increase the helpfulness of the error message.


trieloff:
Do you have details for the plan to remove `hlx init`?

Issue #94, opened 2018-08-16
[static] Create BundledDistributor


trieloff:
The static delivery should follow the `Distributor` interface for all delivery methods, so that we don't have special case handling for the bundled/openwhisk distributor.

tripodsan:
currently this is a bit tricky as it needs to modify the default action.

we might remove the static files from the default action and create a specific 'static' action for that.

Issue #93, opened 2018-08-16
[strains.yaml] align configuration options for static content


trieloff:
Instead of having `githubStatic`, `s3Static`, etc. we should set the configuration as follows:

```yaml
static:
  type: github
  repo: …
  owner: …
  ref: …
```

this makes the setting more aligned with the rest of `strains.yaml` (all lowercase) and enables an easy lookup of distributors.

Issue #60, opened 2018-07-31
Static file handler - Azure Blob Storage


tripodsan:
as mentioned in #53, helix should provide a static file handler that stores the files in Azure Blob Storage.

- [ ] provide configuration option for azure (auth, etc)
- [ ] provide _deployer_ that uploads the static files to azure
- [ ] provide fastly config for Azure access

tripodsan:
don't do for now.

Issue #58, opened 2018-07-30
[hlx deploy] use .zip action for easier extension by default


tripodsan:
For extending the openwhisk actions with additional scripts or content, it would be better to deploy a .zip package instead of the raw script.js.

this is required for #53

PR #112, opened 2018-08-22
WIP: Unified Static Delivery

### LINKED ISSUE #95: [static] redirect to cachable URL
Static resources should have URLs that are unique per deployment. We deliver all static resources with long-lived `Cache-Control` headers, so we need to make sure that requests that point to the generic URL will get redirected to a stable URL that includes for instance the `ref`

trieloff:
There are a number of things that need to be done in this PR. As this is a work in progress, consider this the roadmap

## High-Level

- [ ] use project-helix#266 to deliver all static assets (this fixes #95)
- [ ] remove other static delivery options (bundled, githubStatic)
- [ ] obsolete #54 #55 #56 #57 #60 #94 

## In `hlx deploy`

- [ ] don't package all source files (risk of hitting 1 MB limit) - reverts #59 #58 
- [ ] add information from current remote to `static.{owner|repo|ref}` in `strains.yaml`
- [ ] use `static` as the common configuration for all static stuff: #93 

## In `hlx strain`

- [ ] use parameters from `static` for `static_` edge dictionaries
- [ ] fall back with current `remote` when no `static` configuration can be found (only for default)

## In `helix.vcl`

- [ ] intercept all requests to `/__HLX.*DIST__/` and forward them to static action
- [ ] for `.html` files, use dynamic actions first, fall back to static on 404
- [ ] for all other files, use static action (project-helix#266) first, fall back to dynamic action
- [ ] surpress `X-Static` debug response header

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@adobe/helix-cli",
-  "version": "0.3.1",
+  "version": "0.4.0",
   "description": "Project Helix CLI",
   "main": "index.js",
   "bin": {


tripodsan:
further considerations:
- the dynamic approach can yield unexpected results and is difficult to debug
- the lifecycle of the openwhisk containers is out of our control. i.e. cache hits are very probably and yield to long delivery times
- I would limit static delivery only to `/dist/*` files. there is no need to consider other files as well (if the set of files is built during deploy time)

My counter approach would be:

Highlevel
----------
- [ ] all static files are available (as-is or generated) during deploy time
- [ ] use a pluggable mechanism to deliver static files depending on use case (bundled < 1mb, github, s3, ...)
- [ ] static file source can be decided in edge based on dictionary or url (i.e. /dist/*)

hlx deploy
----------
- [ ] offer different packagers (bundled, github, s3, azure, etc)
- [ ] packagers can signal error if preconditions are not met (eg bundled < 1mb)
- [ ] add static source information to strains.yml 
- [ ] use `static` as the common configuration for all static stuff: #93

hlx strain
---------
-  [ ] use parameters from `static` for edge dictionaries

hlx vlc
--------
- [] intercept all requests to `/dist/` and forward them to static action

tripodsan:
see https://github.com/adobe/project-helix/issues/287

layouts/fastly/helix.vcl
```diff
@@ -271,6 +273,21 @@ sub vcl_recv {
       set req.backend = F_GitHub;
       set req.url = "/" + var.owner + "/" + var.repo + "/" + var.ref + "/" + re.group.1;
     }
+  } elseif (!req.http.Fastly-FF && req.http.Fastly-SSL && req.url.path ~ "__HLX\/([^\/]+)\/([^\/]+)\/([^\/]+)\/([^\/]+)\/(.*)\/DIST__\/(.*)") {
+    # This is the GitHub/OpenWhisk bundler. In this branch of the if statment the __HLX…DIST__ indicates that
```
tripodsan: where is `__HLX`  and `DIST__` set? where does it come from ?

trieloff: From here: https://github.com/adobe/project-helix/pull/266/files#diff-eaa78a59d4b66e15b887023e5b31b4a4R108

I'm using this pattern because it can be used anywhere in the URL, not just in the root path. For instance if you are defining strains based on the path, `/dist` (or `/static`) won't work. These URLs will most frequently be used for assets that are linked from an entry file, so as a developer you don't see them very often.

tripodsan: iirgs...this is very ugly. another reason not to like this approach :-)

trieloff: I've added two items:

> - [ ] hide parts of the URL that can be retrieved from the strain (
https://github.com/adobe/helix-cli/pull/112#discussion_r212419875)
> - [ ] make `__HLX…DIST__` a random string (
https://github.com/adobe/helix-cli/pull/112#discussion_r212419875)

tripodsan: > hide parts of the URL that can be retrieved from the strain
so, basically, just use `/dist` :)

trieloff: More like `$dist`, where `$dist` does not start with a slash so that it works everywhere.

tripodsan: >  does not start with a slash so that it works everywhere.

why would that be required?

trieloff: If your strain is mapped to a path instead of a subdomain going back to `/dist` would reset the strain. By enabling any root path for the `dist` directory you have more flexibility.

trieloff:
I hit a complication while implementing `allow`/`deny` functionality in VCL: regular expression matches in VCL have to string literals, i.e. they can't be loaded from a header or edge dictionary. This means I'll have to delegate the blocking to the action. This also prevents us from using the "VCL + GitHub only" route.

tripodsan:
> This also prevents us from using the "VCL + GitHub only" route.

why? you can generate the blacklist when generating the VLC

tripodsan:
..or we could agree that all static delivered content needs to reside below the dist/docroot/webroot/static/... directory. this would also be very intuitive and similar to how other frameworks work. after all, you can also drive gh-pages from a directory of the main repository.

trieloff:
> -1 we should not remove the support for creating the helix-static branch (i.e. the github distributor).

I can see your dislike for committing back into the same branch, but I'd say this is a) personal preference b) easily implemented in a CI script 3) nothing that warrants the overhead of a separate distributor.

tripodsan:
> I think all mappings that end up in file://…$X -> https://…$Y where $X!=$Y have the downside of being confusing. Things should not be forced to look different in the FS-world and the HTTP-world.

yes. one precondition for this, is that the `docroot` (either manually put there, or via build script) SHOULD be the web root. the `/dist/*` regexp in fastly, is just an optimazation.

> For static -> dist we already know this is a bad idea. But docroot -> / is not any better, and it won't matter if we call it docroot, static, webroot, wwwdata, htdocs. We are breaking continuity and only for the sake of making our implementation a bit easier.

I disagree. The likelyhood that your code repository root has any meaningful files for the static delivery is slim. in your logic, the generated clientside files (less -> css) would need to be put into your code repo root as well...

I say:

```
myproject
├── .hlx/
│   └── build/
│       └── html.js
├── README.md
├── dist/
│   ├── favicon.3453ff.ico
│   └── styles.13acfe.css
├── helix-config.yaml
└── src/
    ├── favicon.ico
    ├── html.htl
    └── styles.less
```

where `hlx build` compiles the `.htl` (into `.hlx/build` and also processes the `less` and other referenced files and puts them into `./dist`.

trieloff:
@tripodsan and then your `const docroot = 'dist'`?

trieloff:
We have `content.root` option, why not a `static.root` option. We can then bicker about the default setting ;)

trieloff:
@tripodsan `static.root` is now a configuration parameter, including documentation. Good to merge from your end?

tripodsan:
> Good to merge from your end?

hmm..no. we still have the __HLX magic in the vlc. where is this generated?
also, please remove the codeload logic from the static.js. this is not needed anymore.
and I tihnk https://github.com/adobe/helix-cli/issues/144 would be a good approach.

trieloff:
I think it will be best to discuss this next week in person. 

- `__HLX` magic is injected only when you turn on Parcel processing (which is optional)
- As I said before, lets see how well or badly the codeload/parcel works before we decide against it

trieloff:
I've removed the `__HLX`and Parcel magic and will open a separate PR for this.

Issue #116, opened 2018-08-23
use native ES6 Promise


stefan-guggisberg:
node.js provides native support for promises since v0.12.18. there's no need to use 3rd party implementations such as `bluebird`. they're an added dependency (which needs to be maintained by greenkeeper, snyk etc.) and they often provide non-standard api's (not covered by the spec).

PR #117, opened 2018-08-23
use native ES6 Promise

stefan-guggisberg:


trieloff:
Do we have to replace `request-promise` with `request-promise-native` to make this work?

stefan-guggisberg:
> Do we have to replace request-promise with request-promise-native to make this work?

good point. ideally yes, but removing the direct `bluebird` dependency doesn't break `request-promise` which includes it's own `bluebird` dependency. 

i'll create another PR.

stefan-guggisberg:
i've updated this PR (replace request-promise with request-promise-native(

trieloff:
Fixed #116

Issue #131, opened 2018-08-31
Align with new Global Scope


kptdobe:
Corresponding issue for https://github.com/adobe/hypermedia-pipeline/issues/7

tripodsan:
is this different than #126  ?

kptdobe:
Arff... apparently not. For whatever reason, I do not watch certain repos thus I did not see it.

Issue #126, opened 2018-08-30
enhance pipeline with external request parameters


tripodsan:
for solving https://github.com/adobe/petridish/issues/46 also for the pipeline, we need:

- [ ] move the magic parameter handling from ow-log-wrapper to cli https://github.com/adobe/openwhisk-loggly-wrapper/issues/6
- [ ] adjust the `OutputTemplate.js` to implement the logic from ow-log-wrapper
- [x] instruct fastly to generate a `req` request parameter that contains the original request

trieloff:
Careful: every request parameter and Cookie is a potential cache-buster. Adding external request parameters to hypermedia-pipeline is easy, but for our VCL we should have a whitelist approach where parameters not on the list get ignored.

tripodsan:
> but for our VCL we should have a whitelist approach where parameters not on the list get ignored.

- how would the developer control this?
- is it at all possible to create an encoded string that contains the external request?

trieloff:
> how would the developer control this?

Another item in the strain config going into an edge dictionary.

> is it at all possible to create an encoded string that contains the external request?

I think so: 
* https://docs.fastly.com/vcl/functions/querystring-filter-except/ allows us to filter the original query string on a whitelist
* https://docs.fastly.com/vcl/functions/urlencode/ allows us to put it safely into another request parameter

kptdobe:
In VCL, encoding the incoming request in JSON to pass it as a request parameter to invoke this action will be a nightmare to handle...
The JSON should be aligned with the one in Petridish: see https://github.com/adobe/petridish/blob/master/src/RequestContext.js#L98-L105

```
{
      url: req.url,
      resourcePath: var.dir +var.name?,
      path: var.dir?,
      selector: var.selector,
      extension: var.extension,
      method: req.request,
      headers: req.http.header?,
      params: req.params?,
}
```

kptdobe:
@trieloff answered:

It should not be too difficult, if we can accept a bit of transforming/wrapping in the action:

- pass it as parameters: `url`, `path`, `selector`, `extension`, `method`
- passed in __ow_headers: `headers`
- passed with a new, url-encoded params parameter: `params` (see above)

The only remainder is `resourcePath`, which should not be set outside the pipeline, but inside the pipeline (as this is where fetching happens). I'd say petridish is wrong here.

trieloff:
[...]

PR #132, opened 2018-08-31
Align with new Global Scope

kptdobe:
PR for #131

### DIFF
--- a/src/parcel/OutputTemplate.js
+++ b/src/parcel/OutputTemplate.js
@@ -22,22 +22,22 @@ function wrap(main) {
   const _isFunction = (fn) => !!(fn && fn.constructor && fn.call && fn.apply);
 
   // this gets called by openwhisk
-  return function wrapped(params, secrets = {}, logger) {
-    const runthis = (p, s, l) => {
-      const next = (p, s, l) => {
+  return function wrapped(payload, action = {}, logger) {
+    const runthis = (p, a, l) => {
+      const next = (p, a, l) => {
         function cont(next) {
-          const config  = Object.assign({}, s, { logger: l });
+          const config  = Object.assign({}, a, { logger: l });
           const ret = pre(p, config);
           if (ret && _isFunction(ret.then)) {
-            return ret.then((pp) => next(pp || p, s, l));
+            return ret.then((pp) => next(pp || p, a, l));
           }
-          return next(ret || p, s, l);
+          return next(ret || p, a, l);
         }
         return cont(main).then(resobj => ({ response: resobj }));
       };
-      return pipe(next, p, s, l);
+      return pipe(next, p, a, l);
     };
-    return owwrapper(runthis, params, secrets, logger);
+    return owwrapper(runthis, payload, action, logger);
   };
 }
 


src/parcel/OutputTemplate.js
```diff
@@ -22,22 +22,22 @@ function wrap(main) {
...
-          return next(ret || p, s, l);
+          return next(ret || p, a, l);
         }
         return cont(main).then(resobj => ({ response: resobj }));
       };
-      return pipe(next, p, s, l);
+      return pipe(next, p, a, l);
     };
-    return owwrapper(runthis, params, secrets, logger);
+    return owwrapper(runthis, payload, action, logger);
```
tripodsan: no. here's it's still `params` and `secrets`

src/parcel/OutputTemplate.js
```diff
@@ -22,22 +22,22 @@ function wrap(main) {
   const _isFunction = (fn) => !!(fn && fn.constructor && fn.call && fn.apply);
 
   // this gets called by openwhisk
-  return function wrapped(params, secrets = {}, logger) {
-    const runthis = (p, s, l) => {
-      const next = (p, s, l) => {
+  return function wrapped(payload, action = {}, logger) {
```
tripodsan: no. here's it's still `params` and `secrets`

tripodsan:
I think we need to include more changes, when we move the owrapper stuff over.

kptdobe:
In VCL, encoding the incoming request in JSON to pass it as a request parameter to invoke this action will be a nightmare to handle...
The JSON should be aligned with the one in Petridish: see https://github.com/adobe/petridish/blob/master/src/RequestContext.js#L98-L105

```
{
      url: req.url,
      resourcePath: <computed out of url>,
      path: <computed out of url>,
      selector: <computed out of url>,
      extension: <computed out of url>,
      method: req.method?,
      headers: req.headers?,
      params: req.params?,
}
```

trieloff:
It should not be too difficult, if we can accept a bit of transforming/wrapping in the action:

- pass it as parameters: `url`, `path`, `selector`, `extension`, `method`
- passed in `__ow_headers`: `headers`
- passed with a new, url-encoded `params` parameter: `params` (see https://github.com/adobe/helix-cli/issues/126#issuecomment-417627335)

The only remainder is `resourcePath`, which should not be set outside the pipeline, but inside the pipeline (as this is where fetching happens). I'd say petridish is wrong here.

kptdobe:
I definitively prefer the `req` parameter idea which separates clearly the action parameters and the original request ones. 
`_ow_headers`: that's the action incoming headers, no ? or at least they have been tweaked by the VLC on Fastly. Might be fine though...

But on the other hand, when I see the amount of VCL effort to simply dump those parameters into one JSON encoded string... 

I found https://github.com/fastly/vcl-json-generate/blob/master/files/json_generate.vcl to inspire from (we would need something similar but not change the req object). But some "JSON utils" would overload the amount of VCL we need.

kptdobe:
Now that I understood what https://github.com/adobe/helix-cli/issues/126 is about, let's move the VCL handling discussion there.

trieloff:
@kptdobe can you take a look at the failing test?

kptdobe:
Yes, on it!

kptdobe:
The `pre.js` code of the projects will need to be adjusted (when requiring owner, repo, path...).
While fixing the tests, I realized that the interesting part of these changes is that you cannot expose some of the internals (like owner, repo or resource path - path to md) without requiring some code: the `it` object in the HTL does not contain those info anymore by default. If you want them here, you need to copy them from `action` to `payload` which I think is good (conscious exposure of "internals")

kptdobe:
The tests work locally but cannot be working in the ci, code relies on the following changes:
- https://github.com/adobe/openwhisk-loggly-wrapper/pull/8
- https://github.com/adobe/hypermedia-pipeline/pull/35

We should merge those first, this will unblock this one.

Issue #144, opened 2018-09-04
generate manifest.json with information of the static files


tripodsan:
since the list of static files is know during deploy time, it can be included into the bundled distributor.

trieloff:
I'd suggest we remove the bundled distributor entirely, given that we have a good way of using GitHub as the distributor.

tripodsan:
this is only a pre-step before we can merge the static-serving logic. As I mentioned somewhere else (can't remember where :-) Having the static delivery redirector (disty) bundled together with the html.js action saves us 1 action. so we still can have a disty per strain, but the config in the VLC is much easier, as we only need to handle 1 case.

tripodsan:
Ok, let's just generate the `manifest.json` for now. I think it could serve as a good source of information for any _static deployment_ mechanism. I will also include the content-hash of the files. eg:

```json
{
  "styles.css": {
      size: 1234,
      hash: "..."
  },
  "favicon.ico": {
  ...
  }
}
```

PR #151, opened 2018-09-05
Bundled with manifest

tripodsan:
see #144


tripodsan:
can we store the manifest in the VLC instead?

tripodsan:
will only create manfest

trieloff:
I'll have to think about it. The manifest can be helpful for dealing with big files.

PR #153, opened 2018-09-05
Edge side implementation of Param whitelist

trieloff:
This is the edge-side implementation for #126: it adds everything on the edge and cli-side to enable passing a parameter `params` that has the names and values of all whitelisted parameters in a URL encoded string. 

On the action-side, you would need to take the value of `params.params` and use `querystring.parse` (https://nodejs.org/api/querystring.html#querystring_querystring_parse_str_sep_eq_options) to get a proper object.

I'm not a 100% sure where you would put that, @kptdobe @tripodsan – therefore only a partial solution to #126

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,31 @@ $ hlx strain --fastly-auth <key> --fastly-namespace <serviceid>
 💀  Purged entire cache
 ```
 
+### Passing Request Parameters
+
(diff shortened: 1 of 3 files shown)


trieloff:
Merging with adobe/hypermedia-pipeline#45

Issue #172, opened 2018-09-10
binary distribution permissions in /usr/local/bin


davidnuescheler:
after the install of the binary distribution it seems the permissions are set to `700` 

```
-rwx------  1 uncled  admin  78706432 Aug 17 18:25 /usr/local/bin/hlx
```

instead of arguably `755` or whatever `umask` suggests.

stefan-guggisberg:
fixed via #185

Issue #170, opened 2018-09-10
dangling link in /usr/local/bin/ prevents install binary distribution


davidnuescheler:
if a `npm install -g @adobe/helix-cli` leaves a dangling symlink as a result of https://github.com/adobe/git-server/issues/13 (which is problematic to begin with) the install of the binary distribution fails.

stefan-guggisberg:
fixed via #185

PR #185, opened 2018-09-11
added permissions and improved symlink handling

davidnuescheler:


davidnuescheler:
should resolve #172 and #170

PR #200, opened 2018-09-18
Use latest parcel

### LINKED ISSUE #180: hash-key of parcel bundles is not environment independent
Parcel uses the full path of the files to create the hash-key. Since the generated static files in `dist` need to be committed to git, running the _project_ on a different environment might produce different hash-keys.

**Suggest:**

compute hash to the relative path to the current directory.

tripodsan:
Changes
- [x] use parcel 1.10.0 (waiting for release)
- [x] hash-key of parcel bundles is not environment independent (Fixes #180)

### DIFF
--- a/test/testBuildCmd.js
+++ b/test/testBuildCmd.js
@@ -15,7 +15,7 @@
 const path = require('path');
 const fs = require('fs-extra');
 const assert = require('assert');
-const { createTestRoot, assertFile, assertFileMissing } = require('./utils.js');
+const { createTestRoot, assertFile } = require('./utils.js');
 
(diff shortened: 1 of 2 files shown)


trieloff:
@tripodsan have you seen/tried the new Parcel beta? https://github.com/parcel-bundler/parcel/releases/tag/v1.10.0-beta.1 (mentioned here: https://github.com/parcel-bundler/parcel/pull/1690#issuecomment-421708889)

tripodsan:
> have you seen/tried the new Parcel beta?

@trieloff did you look at my changed `package.json` at all ? :-)

Issue #80, opened 2018-08-14
[hlx up] use own file watcher instead of parcels


tripodsan:
- better control on when to build
- better output options
- coordinate with static file handling
- circumvent hanging workers-bug in parcel

PR #221, opened 2018-10-02
[hlx up] use own file watcher instead of parcels #80

tripodsan:
see #80 

this PR disables the parcel watcher to recompilation. and restarts the bundler every time a files was modified, added or deleted in the `src` directory. 
it also restarts the petridish server, when the `helix-config.yaml` has changed.


tripodsan:
btw, the reduction of the test coverage is due to the problems we have running the content-invalidation tests; due to some hard to debug hanging processes from parcel.

PR #238, opened 2018-10-04
[hxl demo] does not provide a default .gitignore (released package on…

### LINKED ISSUE #228: [hxl demo] does not provide a default .gitignore
```
hlx demo test
cd test
hlx build
git status
```

`.hlx` is shown as untracked and will be disturbing the next `hlx deploy` (uncommitted files).

Here is a decent .gitignore file to be included:

```
.hlx/cache
.hlx/build
node_modules/
logs
```

tripodsan:
fixes #228


kptdobe:
I know know why the `.cache/` from my `.gitignore` is coming from...  https://github.com/adobe/helix-cli/blob/2e6bc0c64b3321c47258d548a16012a91356a6ca/demos/full/_gitignore#L2 ;)

tripodsan:
> My only feedback would be: why the tests did not detect this was not working anymore ? Maybe we should had a test to more sure this does not happen anymore.

because it only happens with the published package. i.e. during npm publish, the .gitignores are removed.

PR #243, opened 2018-10-04
[VCL] Compute path properly and use it where needed

kptdobe:

## Related Issues
PR for https://github.com/adobe/helix-cli/issues/240

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -388,27 +388,16 @@ sub vcl_recv {
         }
       }
 
-      // compute full resource path with appropriate number of "/"
-      if (var.dir == "/") {
-        set var.path = "/" + req.url.basename;
-      } else {
-        set var.path = var.dir + "/" + req.url.basename;
-      }
-
+      set var.path = var.dir + "/" + req.url.basename;
       set req.url = "/" + var.owner + "/" + var.repo + "/" + var.ref + var.path + "?" + req.url.qs;
     } else {
-
-      // compute full resource path with appropriate number of "/"
-      if (var.dir == "/") {
-        set var.path = "/" + var.name + ".md";
-      } else {
-        set var.path = var.dir + "/" + var.name + ".md";
-      }
-
+      set var.path = var.dir + "/" + var.name + ".md";
       # Invoke OpenWhisk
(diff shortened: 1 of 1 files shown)


kptdobe:
I would like to write some tests to cover this case. I could not find a similar (integration) test to "inspire" from. @tripodsan @trieloff is there a test somewhere that would test the request output (like using the X-Debug and test the expected headers) ?

tripodsan:
I think the error could already happen earlier, if you have a root path that ends with a slash.
i.e. in:

https://github.com/adobe/helix-cli/blob/v0.7.12/layouts/fastly/helix.vcl#L345

I think in the end, we just need to sanitize all urls and paths in the action before it goes into the pipeline.

tripodsan:
wouldn't it be easier to sanitize the url and path with a regexp ?

```
set req.url = regsuball(req.url, "/+", "/");
```

layouts/fastly/helix.vcl
```diff
@@ -387,13 +387,18 @@ sub vcl_recv {
...
+      set req.url = "/" + var.owner + "/" + var.repo + "/" + var.ref + var.path + "?" + req.url.qs;
     } else {
+      set var.path = var.dir + "/" + var.name + ".md";
       # Invoke OpenWhisk
-      set req.url = "/api/v1/web" + var.action + "?owner=" + var.owner + "&repo=" + var.repo + "&ref=" + var.ref + "&path=" + var.dir + "/" + var.name + ".md" + "&selector=" + var.selector + "&extension=" + req.url.ext + "&branch=" + var.branch + "&strain=" + var.strain + "&GITHUB_KEY=" + table.lookup(secrets, "GITHUB_TOKEN");
+      set req.url = "/api/v1/web" + var.action + "?owner=" + var.owner + "&repo=" + var.repo + "&ref=" + var.ref + "&path=" + var.path + "&selector=" + var.selector + "&extension=" + req.url.ext + "&branch=" + var.branch + "&strain=" + var.strain + "&GITHUB_KEY=" + table.lookup(secrets, "GITHUB_TOKEN");
     }
   }
 
+  set req.url = regsuball(req.url, "/+", "/");
```
trieloff: This can become a problem when we introduce URL parameters, where `//` is a completely acceptable value. Probably better to just sanitize `var.path`.

Issue #253, opened 2018-10-08
Fix to default image TTL


ejthurgo:
**Description**
A bug was identified in the Fastly IO Boilerplate VCL which can result in poor cache performance.
When no cache control headers are passed up from the origin, we fall-back to default TTLs specified in the IO boilerplate. For image types, this was resulting in different TTLs at the edge and shield and in turn, increased cache misses from edge to shield.

**To Reproduce**
1. Origin returns no `cache-control` header for supported image file type

**Expected behavior**
Deploy consistent TTLs across edge and shield.

**Screenshots**
None.

**Version:**
0.7.12

**Additional context**
None

PR #254, opened 2018-10-08
Fixes Reduced CHR when applying default image TTLs

ejthurgo:
Recently uncovered an issue with the Fastly IO Boilerplate which can
result in reduced CHR when the origin doesn’t pass up caching headers
and we fallback to default TTLs.
Issue would exhibit itself by decreased TTL at the edge resulting in
increased miss-rate to the shield.

Resolves issue: https://github.com/adobe/helix-cli/issues/253


trieloff:
Thank you @ejthurgo

Issue #247, opened 2018-10-07
Add a "X-Version" header that exposes the Fastly configuration version


kptdobe:
**Is your feature request related to a problem? Please describe.**
After a `hlx strain`, the cache is purged and the first hit is sometimes really slow but sometimes it is surprisingly fast: this gives a lot of doubts and uncertainty knowing if the request was made before the purge fully happened and if the content has really been updated. The fact is that I often do not know if my change has been taken into account thus I need to `hlx strain` again.

**Describe the solution you'd like**
My proposal is to add a X-Version header that contains the Fastly version number. It could be "injected" somehow during the `hlx strain` process. Like this, I am sure which version generated the html delivered.

cc @trieloff @tripodsan

kptdobe:
I takes sometimes multiple `build / deploy / strain` to get the latest change (htl in my current case) on the "publish" website. There is nothing wrong somewhere and hopefully, being able to determine which VCL version + action "version" would be necessary.

kptdobe:
Arff.. the pb in the current case is that I want to check before committing, thus I added `--dirty` which changes the action name thus I should have updated my strains... something else to improve...

kptdobe:
The main issue: this cannot live only in Fastly because activating a new version is independent from purging the cache. You can cache A.html with v12, activate v13, do not purge the cache, the current active version is not the one that "created" A.html.

trieloff:
`hlx publish` runs a complete purge at the end, so you are unlikely to experience the issue you've outlined.

kptdobe:
Implemented with https://github.com/adobe/helix-cli/commit/00e01d2f9f514b04aa7d659be7e4077718a19222

tripodsan:
I disagree:
- using the VCL version retrieved during hlx publish, is not accurate, if someone modifies the VLC afterwards via API or UI. if we only set it in hlx publish, we might provide a wrong number, which will be even more confusing. if it doesn't match the one that is active on fastly.

please reconsider to set the VLC version using the `req.vcl` variable, as proposed by fastly support.

/cc @trieloff

kptdobe:
You cannot modify the VCL of an active version, you need to clone it which will create a new version number, in which case, the version will be wrong, I agree. 
Actually, we could have both, this will tell us immediately if someone did not use `hlx publish` to generate the config! Even better.

tripodsan:
[...]

PR #255, opened 2018-10-08
Add a "X-Version" header that exposes the Fastly configuration version

kptdobe:
PR for https://github.com/adobe/helix-cli/issues/247


kptdobe:
Now how to make it clean: 

* if we agree on the approach, the only thing I do not like is to generate the X-Version line in the `strains.vcl` file.
* 2 options:
  1. create a new file for other future generated stuff 
  2. rename the `strains.vcl` to `dynamic.vcl` which contains everything that need to be generated  and re-format the code a little bit to aggregate the various dynamic parts in various functions.

I vote for option 2. Simple and easy.

@trieloff @tripodsan WDYT ?

trieloff:
2 sounds good. Here are a couple of things that would look good:
- VCL version number
- Git Commit SHA
- HLX CLI Version Number

layouts/fastly/helix.vcl
```diff
@@ -28,7 +28,7 @@ sub hlx_strain {
     set req.http.X-Strain = "default";
 
     # run custom strain resolution
-    include "strains.vcl";
+    include "dynamic.vcl";
```
trieloff: I would probably split the two includes. Keep one `strains.vcl` and have an entirely new `dynamic.vcl`

kptdobe:
Added some few things:
- kept the `strains.vcl` file and added a `dynamic.vcl` file
- slightly changed the publish process to fix a bug: `helix.vcl` cannot be deployed if `strains.vcl` or `dynamic.vcl` does not exist (include is checked and fails)! This would block the really first deployment...
- `x-version` looks like this: `20 | 0.8.1-pre.4 | 46f590ef25ab5beb060d537f5fb2db91c1e61bec`
- format is `<fastly config version> | <hlx cli version> | <current folder version (project code)>`
- I changed to output X-Version all the time (no X-Debug required). I think this is really useful until things get stable and helps really being sure the content viewed in the browser is generated with latest code. We can change that back when things are rolling.

> VCL version number

What do you mean ? We do not have such a version number. But the cli version should give you a hint.

Still TODO: add the tests for all the changes

trieloff:
VCL version number is covered with `<fastly config version>`.

Not super thrilled about having the header in by default, as our headers tend to get chattier over time and then we have to consolidate again.

trieloff:
Maybe we could have a "weak debug" option, that we turn on when the user agent is `curl`, etc.

kptdobe:
Arff... not so simple! The `dynamic.vcl` is... dynamic thus cannot predict (record and replay) the POST request to upload the vcl.

kptdobe:
> Not super thrilled about having the header in by default, as our headers tend to get chattier over time and then we have to consolidate again.

@trieloff I clearly see that doing the same request with `X-Debug` and without triggers a new action call which means the 2 requests are cached independently. This also meaning we lose the information with which versions the non-debug request has been initially generated with when we want to debug a production site.
I am just saying we might need it with the first site(s) going on production for "debugging". Once this is stable or if this is useless, we can easily get rid of it. I created https://github.com/adobe/helix-cli/issues/264 to track the removal.

tripodsan:
maybe we should ask fastly to put the configuration version somewhere in their https://docs.fastly.com/vcl/variables/.

tripodsan:
message from fastly:

> Hi Tobias,
>
> I think the req.vcl variable is the best one for your purposes, below is a quick definition:
>
> req.vcl - STRING - Read Only - Available in vcl_recv
> The version of the customer VCL currently loaded. It will take the form [service_id].> [version_number]_[generation]-[hash].
>
> I've also created a fiddle for you to see what the output of this looks like:  https://fiddle.fastlydemo.net/fiddle/e7282e34

tripodsan:
I quickly tested it, and it works. for example, I get:

```
72jihuXMkAS9sNuyA7SFpL.20_7-47c7204417d30617822e1eb6543cab77
```

kptdobe:
Interesting, `req.vcl` is not documented...

Since @trieloff wanted to have more than the config version (cli version and current folder last commit sha), the x-version format is `<fastly config version> | <hlx cli version> | <current folder revision (project code)>`.
We anyway need to compute the `X-Version` dynamically to append the last 2. And since gathering the version is anyway the first thing the publish command is doing, I do not see the need to use this variable for now. If we would do everything in VCL, yes, but we compute everything in JS anyway.

Issue #259, opened 2018-10-09
[hxl strain] fires an exception at the end of the process (sometimes)


kptdobe:
**To Reproduce**
Just run `hlx strain` and sometimes you get output logs:

```
All dicts have been updated.
🚀  Activated latest version, version 69 is live
✅  VCL helix.vcl has been updated
Unable to update VCL helix.vcl
422 - {"msg":"Version locked","detail":"Version '69' for service '3t7g8apsfQlFyvuHIQn9e8' is locked"}
(node:83077) UnhandledPromiseRejectionWarning: Error: Unable to update VCL helix.vcl
    at request.then.catch (/Users/alex/work/dev/helix/project-helix/modules/helix-cli/src/publish.cmd.js:495:15)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:182:7)
(node:83077) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:83077) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
💀  Purged entire cache
✅  All strains have been published and are online.
```

**Version:**
run: `$ hlx --version`
0.8.1-pre.2

trieloff:
There seems to be a hard-to-reproduce condition where the update of the VCL sometimes happens after the version gets published.

kptdobe:
Yes, but the reason and the fix is easy ;)
Working on the PR...

PR #260, opened 2018-10-09
SetVCL is async, need to wait for result

kptdobe:
PR for https://github.com/adobe/helix-cli/issues/259


kptdobe:
For whatever reason, the requests do not go through `node-replay`... need to understand what happens.

kptdobe:
I added a breakpoint in the node-replay patch of the request method and it never goes in there. Not sure why it does not work on my env and on the CI (or even how this would have ever worked...)

tripodsan:
I fixed the tests. Probably some API changed and then replay couldn't find the recorded responses.

trieloff:
@tripodsan did you re-record with a fresh service config? It looks like the test coverage has gone down, and this is typically the coverage of the "create everything from scratch" part.

kptdobe:
Yes he did, the service Id has changed.
I checked the coverage report and it says that some lines have been changed in `up.cmd.js` which not true. Very strange.

tripodsan:
> @tripodsan did you re-record with a fresh service config? It looks like the test coverage has gone down, and this is typically the coverage of the "create everything from scratch" part.

yes. I _improved_ the error handling when setting the VLC. I think it might have more lines now. But since we don't have any tests, that cover the error handling, the coverage went down.

see https://github.com/adobe/helix-cli/pull/260/files#diff-64a55b500b78cf4831b3bf2ed1f97953

trieloff:
We could probably copy the fixtures directory, edit some of the responses manually and replay the tests with broken responses to test the error case.

PR #265, opened 2018-10-12
Use req.vcl to determine the Fastly config version

kptdobe:
PR for https://github.com/adobe/helix-cli/issues/247

### DIFF
--- a/src/publish.cmd.js
+++ b/src/publish.cmd.js
@@ -443,12 +443,12 @@ class PublishCommand {
     return retvcl;
   }
 
-  static getVersionVCL(configVersion, cliVersion, revision) {
+  static getXVersionExtensionVCL(configVersion, cliVersion, revision) {
     let retvcl = '# This section handles the strain resolution\n';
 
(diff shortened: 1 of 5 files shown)


src/publish.cmd.js
```diff
@@ -443,12 +443,12 @@ class PublishCommand {
     return retvcl;
   }
 
-  static getVersionVCL(configVersion, cliVersion, revision) {
+  static getXVersionExtensionVCL(configVersion, cliVersion, revision) {
     let retvcl = '# This section handles the strain resolution\n';
 
-    const version = `${configVersion} | ${cliVersion} | ${revision}`;
+    const version = ` | ${configVersion} | ${cliVersion} | ${revision}`;
```
tripodsan: @kptdobe, @trieloff 
I think we should use default http header semantics with subfields, eg:

```
X-Version: 2; src=2; cli=0.8.1-pre.4; hash=6bd3c895a62e4909eb2232e0ed7a984f5c4b1b69
```

PR #272, opened 2018-10-17
new Param whitelist

trieloff:
- [x] this is the definitive fix for #126

## Related Issues
- #126 
- #153


layouts/fastly/helix.vcl
```diff
@@ -397,10 +395,24 @@ sub vcl_recv {
...
+      set req.url = "/api/v1/web" + var.action + 
+        "?owner=" + var.owner + 
+        "&repo=" + var.repo + 
+        "&ref=" + var.ref + 
+        "&path=" + var.path + 
+        "&selector=" + var.selector + 
+        "&extension=" + req.url.ext + 
+        "&branch=" + var.branch + 
+        "&strain=" + var.strain + 
+        "&GITHUB_KEY=" + table.lookup(secrets, "GITHUB_TOKEN") +
```
tripodsan: I'm really not happy to pass the token as request parameter. can we send a POST instead?

trieloff: We can't, but it isn't too much of a problem, as we can pass the token as a default parameter to the action. The `GITHUB_TOKEN` wasn't used anyway, so I've removed the offending line.

Issue #43, opened 2018-07-19
Reconsider strains.yaml location


kptdobe:
`strains.yaml` ends up in the `.hlx` folder. It is generated by hlx, but I would not expect to edit files in `.hlx` folder. We never edit files in the `.git` folder…

kptdobe:
@trieloff says:

I’ve modeled it after CircleCI’s `.circleci/config.yml` but we can change it.
One way of doing it would be to have `helix-dev.yaml` (i.e. `helix-config.yaml`) and `helix-prod.yaml` (i.e.`.hlx/strains.yaml`)

tripodsan:
see https://github.com/adobe/petridish/issues/75

PR #273, opened 2018-10-17
POC: eno config language

trieloff:
While reviewing and thinking about adobe/petridish#76 I looked up the ENO configuration language (@tripodsan I think I mentioned it a while ago, forgot the name, could not remember it when you asked me, and found it again by remembering where I might have seen it).

There is a good introduction to the language at: https://eno-lang.org/introduction/

Some notable features:

- looks like markdown
- does not use indentation in ways that break easily (I'm looking at you, YAML)
- supports copying and deep copying (https://eno-lang.org/advanced/)
- has a good JS API: https://eno-lang.org/javascript/
- has a VS Code Plugin: https://eno-lang.org/plugins/

Now, take a look at the code. I've translated the example from adobe/petridish#76 with some extra that mean:

- we could drop all implicit inheritance logic we have right now in the CLI and the VCL by making all inheritance explicit. When defining a new strain, you either copy from `default` (`## mystrain << default`) or the CLI will throw an error when a required field is missing
- we get stuff like content references (adobe/petridish#76) for free, simply by using statements like `### content < project-helix.io content`, which means the content of the strain is copied from a complex definition elsewhere
- editing config files for local development also gets easier, as you can just define a `local` strain that mixes and matches settings from other strains

Just a very early POC, but it might be worth the consideration when we have to break config file compatibility anyway.

## Related Issues

- #271
- #43 
- adobe/petridish#76


tripodsan:
hmm. funny.

what I don't like is that the configuration language is used for references and extending, which ties it very strongly to the data structures. i.e. if you want to use a different configuration mechanism, you need to express this in another way, which might not be possible.

trieloff:
I'm thinking just the opposite. The reference mechanism is one of the strongest features, even if we would want to use a different notation.

trieloff:
Speaking of a different notation: it looks like YAML has references, too: https://blog.daemonl.com/2016/02/yaml.html but it looks quite ugly (and I don't know if our library supports it)

trieloff:
Looks like js-yaml, which we use supports it: https://github.com/nodeca/js-yaml/blob/master/test/samples-common/construct-merge.yml

trieloff:
@kptdobe what do you think?

kptdobe:
I do not like:
- it is close to markdown, but different and then confusing (# ## ### for the sections look really bad)
- it is a "langage" I have to learn just to setup some few parameters
- the references. It looks great on the first look but what happens when you start having E > D > C > B > A ? How do I know what is in E with all the inheritance and overrides ? Can I debug it ?

Why not using [Config.js](https://github.com/lorenwest/node-config) like every one else ? :)

trieloff:
Ok, I'll take that as an overwhelming no for ENO.

What do you think of the idea of dropping implicit (application-level) inheritance and relying completely on config-file-level inheritance? I fear that we will end up with too many mechanisms that are all slightly incompatible. What we have right now is:

- fallbacks in VCL (first look up the strain, then the default strain, then use the application default)
- fallbacks in the CLI (first look up the strain, then the default strain, then use the application default)
- fallbacks in adobe/petridish#76 (first the strain, then the config-file-wide default)
- plus content references (can't find the issue anymore)

kptdobe:
I do not get it. What would be the replacement for those fallbacks ? As soon as you have a config file, the app has to assume it can be missing / broken / invalid and provide some fallbacks. Otherwise what happens ? error page ?

trieloff:
If you don't have a value for a required setting, we don't deploy. 
If you don't have a value for an optional setting, we use the application-wide default.

If you don't like to repeat yourself in the config file, you use YAML references.

trieloff:
Closing, suggestion rejected.

Issue #250, opened 2018-10-08
Don't set X-Strain Cookie unless required


trieloff:
For most conditions, in particular the `url` condition, setting the `X-Strain` cookie is not needed and can lead to additional issues in development, so it should not be set at all.

As a first step, we can only set the cookie when the `condition` property is set, but skip it when `url` is set.

In a second step, we can add a property `sticky` (off by default) that toggles cookie setting for complex conditions.

trieloff:
@kptdobe this might be a dependency for project-helix.io, if we want to enable the path-based switching between different repository roots.

PR #278, opened 2018-10-18
Non-Sticky Strains

trieloff:
Fix for #250:

the Strain Cookie will only be set if there is an arbitrary `condition`. For `url`-based dispatch, there is no need for sticky strains, because the resolution is predictable and will always be the same for the same URL.


src/publish.cmd.js
```diff
@@ -417,6 +422,10 @@ class PublishCommand {
       return Object.assign({
         condition: `req.http.Host == "${uri.host}"`,
       }, strain);
+    } if (strain.condition && strain.sticky === undefined) {
```
tripodsan: please indent properly

src/publish.cmd.js
```diff
@@ -429,7 +438,10 @@ class PublishCommand {
     const conditions = strains
       .map(PublishCommand.vclConditions)
       .filter(strain => strain.condition)
-      .map(({ condition, name, vcl = '' }) => `if (${condition}) {
+      .map(({
+        condition, name, vcl = '', sticky = false,
```
tripodsan: would be nice to have `sticky = false,` on the next line

Issue #283, opened 2018-10-19
Default CI config cannot be executed ootb


kptdobe:
**To Reproduce**
```
hlx demo test
cd test
hlx deploy
```

Go to the CI url that appears at the end of the deploy execution: you'll see a build failure.

Multiple issues:

- [ ] `git diff` times out
- [ ] `hlx deploy` fails because of local modifications
- [ ] `hlx perf` fails because of missing `calibre-auth`

**Expected behavior**
The initial execution must be smooth and be running end to end. If extra configs are expected, then it must be documented.

**Version:**
run: `$ hlx --version`
0.8.1-pre.8

PR #284, opened 2018-10-19
Default CI config cannot be executed ootb #283

kptdobe:
PR for #283 

- [x] `git diff` times out
- [x] `hlx deploy` fails because of local modifications
- [x] `hlx perf` fails because of missing `calibre-auth`


trieloff:
It looks like `helix-cli` won't read `HLX_CALIBRE_AUTH`, but works with `--calibre-auth`.

Issue #317, opened 2018-11-12
update petridish@2.5.0 to helix-simulator@2.6.1


tripodsan:
the PR in https://github.com/adobe/helix-cli/pull/314 was never run on circleci.
maybe because it originated from a forked repo?

@stefan-guggisberg I created a new PR in #318. can you have a look why it fails on [circleci](https://circleci.com/gh/adobe/helix-cli/1506?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) ?

@trieloff do you know why circleci doesn't run jobs for PRs for forked repos?

stefan-guggisberg:
the reason why it works on `git-server` and `helix-simulator` circleci builds but fails on `helix-cli` is the different docker image versions being used. `helix-cli` uses `circleci/node:8-stretch` (debian 9) whereas the others are using `circleci/node:8` (debian 8).

i've found similar issues related to openssl on debian 9 vs debian 8.

PR #318, opened 2018-11-12
update petridish@2.5.0 to helix-simulator@2.6.1 #317

tripodsan:
see #317


tripodsan:
thanks @stefan-guggisberg. I think I changed it one to get a more modern npm, but we don't need it anymore.

Issue #334, opened 2018-11-19
Pure-JS actions are not wrapped 


tripodsan:
**Is your feature request related to a problem? Please describe.**
as discussed in #328, _pure-js_ actions (eg: `html.js` or `html_dump.js` are currently not transformed/wrapped by parcel, and end up as-is in the openwhisk action. this means the developer needs to write (and understand) the openwhisk web action invocation semantics.

**Describe the solution you'd like**
Ideally, such _pure-js_ actions are automatically wrapped with parcel so that the developer can write them the same way as the `...pre.js` actions.

**Describe alternatives you've considered**
the only alternative is to write a `pre.js` and have an empty `.htl`

trieloff:
@rofe you've been experiencing this.

trieloff:
@tripodsan did the merge of #325 fix this?

tripodsan:
with this fixed, pure action can be written with a simple `main` pipeline function. for example:

**src/xml.js**
```js
module.exports.main = async function main() {
  return {
    body: `<?xml version="1.0" encoding="UTF-8"?>
<note>
  <to>Peach</to>
  <from>Mario</from>
  <heading>Reminder</heading>
  <body>Don't forget me this weekend!</body>
</note>`,
  };
};
```

rofe:
Not sure yet if this is a `helix-cli` or `helix-pipeline` (specifically [#106?](https://github.com/adobe/hypermedia-pipeline/issues/106)) problem, but my example of a pure js action fails in the `validate` action due to an interestingly looking `response` object during step 6 after the `main` function (where I simply return `payload`) got executed:

main function in test_xml.js:
```
async function main(payload, action) {
    return payload;
}
````

payload dump:
```
{
    "content": {
        ...
    },
    "request": {
        ...
    },
    "response": {
        "request": {
            ...
	},
	"content": {
	    ...
        }
    }
}
```
Looks like at some point `response` got confused with `payload` and `request` and `content` objects got duplicated inside...

rofe:
The `response` seems to be merged into `payload` right after the 'main' function. The error message is ```Error: Invalid Context at step 6
data.response should NOT have additional properties, data.response should NOT have additional properties```

trieloff:
@rofe try changing two things:

1. in pipeline: `content.xml` to `response.xml` (also in the schema)
2. in your `main`: `return {xml: payload};

tripodsan:
[...]

PR #337, opened 2018-11-21
Pure-JS actions are not wrapped #334

tripodsan:
see #334

- adds logic to the parcel bundling that triggers wrapping of all non `.pre.js` assets

### todo
- [x] need tests
- [x] need to work for sync and async main() functions. (currently only async)
- [x] how to reject non action js ?

### discussion
the default file filter select all `src/**/*.js` as sources for scripts. we can already filter out the `.pre.js`, but any _helper_ scripts would be difficult to detect.

for example:

```
src
  - html.htl
  - helper.js
  - json.js (requires helper.js)
```

#### possible solutions:
- a. detect if a _helper_ is required by another script, and then reject it as pure-JS script (shaky and might be difficult to implement with current parcel handling)
- b. detect if a script contains a `module.exports.main` and then reject if not (shaky, as a helper also might export a `main`)
- c. only match for known media types: `src/**/*html.js, src/**/*json.js, src/**/*xml.js` etc. might be ok, but is not easy extendable.
- d. only match toplevel directory and put helpers in a sub directory (what about future components) ?
- e. don't deal with this, developer needs to put helper into a directory outside `src` (eg `lib`). problem: file watcher needs to know about this.

for now a combination of b. and e. might be sufficient. 

@trieloff maybe you have and idea?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,21 +9,21 @@
       "resolved": "https://registry.npmjs.org/@adobe/git-server/-/git-server-0.9.10.tgz",
       "integrity": "sha512-TjALKfn4qxskerRkl8k03lI3PQqzB/w9qzreiq8B2qibK+59XK/qsERB5nVZy0dlxKJ0od/SQ/t8LueD+tYg0A==",
       "requires": {
-        "archiver": "3.0.0",
-        "async": "2.6.1",
-        "escape-html": "1.0.3",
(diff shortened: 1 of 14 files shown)


trieloff:
Why don't we start with (c) as the default. You can still override it at the command line and we come up with a more elegant solution when it becomes a recurring issue.

Issue #232, opened 2018-10-04
Change default of auto deploy to `no-auto`


tripodsan:
see #229.

### Suggest:
- change default of `--auto` to `false`

trieloff:
Changing the default is just masking the underlying issues outlined in #229: -1

tripodsan:
The problem is, that for initial development, no one wants auto. it's just another hurdle to get across. 

I think the natural evolution of a project is:
1. completely local (`hlx up`)
2. manual deploy (`hlx deploy --no-auto`)
3. fastly integration (`hlx publish`)
4. CI integration (`hlx deploy`)

I think that is more a problem is, that `hlx deploy --auto` does 2 things:
- it creates the actions
- and sets up circle ci

maybe we should separate this; either by using the helix-bot, which sets up the circleci automatically, or with a separate command `hlx enable-autodeploy`. also the strains config should probably define, which branches/strains are managed by the CI.

ps: For the completely managed case, you don't even need `hlx deploy`; you just commit the code and the helix-bot does the rest.

trieloff:
Ok. Let's change it.

tripodsan:
@rofe do you want to take a stab on this?

rofe:
Ack. I can look into this right after the [go-live](https://github.com/adobe/Adobe-Developer-Site).

PR #355, opened 2018-11-28
Change default of auto deploy to `no-auto` #232

rofe:
Implements #232


rofe:
Checking...

trieloff:
Do you check the opposite? If enabling auto-deploy works?

rofe:
Working on it...

test/testDeployCli.js
```diff
@@ -165,29 +165,32 @@ Authentication is required. You can pass the key via the HLX_WSK_AUTH environmen
...
+        done();
+      })
+      .run(['deploy', '--auto', 'true',
         '--wsk-auth', 'secret-key',
         '--wsk-namespace', 'hlx',
-        '--no-auto',
       ]);
 
-    sinon.assert.calledWith(mockDeploy.withEnableAuto, false);
+    sinon.assert.calledWith(mockDeploy.withEnableAuto, true);
```
tripodsan: those checks are executed too early. you need to either move them into the `onFail()`  case, or even better:
- change the test to be `async` (instead of done)
- `await` for the run
- add a try/catch around it, and check the error there
- add the asserts after the try/catch

rofe: seems correct

Issue #341, opened 2018-11-21
Progress bar for `hlx deploy` is jumpy and skippy


trieloff:
Running `hlx deploy` creates a progress bar that jumps back and forth (from nearly 100% to 80%) and that creates line breaks, resulting in output like this:

```bash
$ hlx deploy --no-auto
[==================================================] analyzing static.js 0.0s
[================================------------------] packaging node_modules/request-promise-native/package.json 0.2
[==================================================] packaging node_modules/lodash/package.json 0.0s
[==================================================]  0.0s

✅  deployment completed
```

tripodsan:
hmm... it should only create 3 progress bars. 1 for analyze, one for packageing and 1 for deployment.

kptdobe:
Closed during backlog grooming session during hackathon. Reopen if needed or open a new one.

PR #361, opened 2018-11-29
Better progress bars

trieloff:
Partial fix for #341: no more line breaks in the deploy progress bar, but we should still try to have only one progress bar per action

### DIFF
--- a/src/deploy.cmd.js
+++ b/src/deploy.cmd.js
@@ -176,11 +176,12 @@ class DeployCommand extends AbstractCommand {
 
     const ticks = {};
     const tick = (message, name) => {
+      const shortname = name.replace(/\/package.json.*/, '').replace(/node_modules\//, '');
       bar.tick({
-        action: name ? `packaging ${name}` : '',
+        action: name ? `packaging ${shortname}` : '',
       });
       if (message) {
-        this.log.log({
+        this.log.maybe({
           progress: true,
           level: 'info',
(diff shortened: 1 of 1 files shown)


src/deploy.cmd.js
```diff
@@ -176,18 +176,24 @@ class DeployCommand extends AbstractCommand {
...
-        this.log.log({
+        this.log.maybe({
           progress: true,
           level: 'info',
           message,
         });
       }
     };
 
+    if (this._dryRun) {
```
tripodsan: no. this is important for testing (also manually). unless we separate `packaging` step from the `deploy` we need this.

src/deploy.cmd.js
```diff
@@ -176,18 +176,24 @@ class DeployCommand extends AbstractCommand {
...
     const ticks = {};
     const tick = (message, name) => {
+      const shortname = name.replace(/\/package.json.*/, '').replace(/node_modules\//, '');
       bar.tick({
-        action: name ? `packaging ${name}` : '',
+        action: name ? `packaging ${shortname}` : '',
       });
       if (message) {
-        this.log.log({
+        this.log.maybe({
```
tripodsan: why?

trieloff: `maybe` only logs when no TTY is attached. I suspected the log message of introducing line breaks.

PR #363, opened 2018-11-30
Pure-JS actions are not wrapped #334

### LINKED ISSUE #334: Pure-JS actions are not wrapped

tripodsan:
fixes #334


tripodsan:
@kptdobe don't know why the smoke test fails. I tested the smoke-test locally, and it works.

kptdobe:
Well... looking at https://circleci.com/gh/adobe/helix-cli/2094, the message is pretty obvious especially right after your change for https://github.com/adobe/parcel-plugin-htl/issues/44.

PR #391, opened 2018-12-14
🐛 fix infinite watch / build loop that only affected me

filmaj:
Turns out the problem was because I have all of my projects under `~/src/`, i.e. `~/src/Adobe-Developer-Site`, `~/src/helix-cli`, etc. So, the first conditional in this function never was true, so the short-circuit didn't work on my machine.

My approach to fix was to remove the process' current working directory from the absolute path of changed filenames passed into the watcher method.

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/up.cmd.js
+++ b/src/up.cmd.js
@@ -68,14 +68,15 @@ class UpCommand extends BuildCommand {
     this._watcher = watch(this.directory, {
       recursive: true,
     }, (eventType, filename) => {
-      if (filename.indexOf('src/') < 0 && filename !== HELIX_CONFIG) {
+      const file = filename.split(process.cwd()).join(''); // only consider paths starting from project root
+      if (file.indexOf('src/') < 0 && file !== HELIX_CONFIG) {
         return;
       }
       // ignore some files
-      if (/(.*\.swx|.*\.swp|.*~)/.test(filename)) {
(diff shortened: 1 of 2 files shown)


src/up.cmd.js
```diff
@@ -68,14 +68,15 @@ class UpCommand extends BuildCommand {
     this._watcher = watch(this.directory, {
       recursive: true,
     }, (eventType, filename) => {
-      if (filename.indexOf('src/') < 0 && filename !== HELIX_CONFIG) {
+      const file = filename.split(process.cwd()).join(''); // only consider paths starting from project root
```
tripodsan: can't you use `this.directory` instead of `cwd`?
or better, `path.relative(this.directory, filename)`

I think the new watcher introduced by @trieloff now returns absolute paths. the previous one, returned relative ones.

filmaj: Good suggestion. Based on the original conditional, if the watcher now returns absolute paths, then without this fix, the check against `HELIX_CONFIG` also won't work.

filmaj:
Another thing I started to try to implement but gave up was writing a test for this. Seems like we only have "integration" tests for this command. I gave up and figured it is not worth adding a heavy integration test for this bug but if y'all feel otherwise let me know and I can spend some more time on that.

Issue #390, opened 2018-12-14
hlx deploy doesn't include local required files into zip


tripodsan:
**Description**
_sometimes_ some of the required module won't end up in the action .zip package, but the packager creates an action instead.

**To Reproduce**
run the 'deploy create correct package' test in the 'problems-with-build-package' branch.

**Expected behavior**
the local file (in this case, the `src/helper.js`) is included in the zip.

Issue #336, opened 2018-11-21
Use canonical action names independent of git origin


tripodsan:
the action name is generated with:

```
`${GitUtils.getRepository()}--${GitUtils.getBranchFlag()}--`,
```

which yields a different name, depending if the use cloned the repo with _ssh_ or _https_. for multi-dev projects this can yield to compatibility problems.

**Suggest**

use format:

```
<hostname>-<owner>-<repo>--<ref>--<scriptname>
```

tripodsan:
as per discussion in https://github.com/adobe/helix-shared/pull/12#issuecomment-448257229, this probably changes to a hash....

PR #392, opened 2018-12-14
hlx deploy doesn't include local required files into zip #390

tripodsan:
see #390


tripodsan:
I need to disable the relevant tests due to: #336,
which makes it impossible to test the generated zip easily.

PR #423, opened 2019-01-03
Fix parcel-bundler regression

koraa:
parcel-bundler 1.11.x moved it's logger library into
it's own repo. This just makes sure we do not automatically
upgrade to that version

## Related Issues

Thanks for contributing!


trieloff:
We're using exact versions instead of version ranges for most things. Can't we just add the new parcel logger dependency?

koraa:
Pulling first would have been helpful, I had a master from november where versions wherent pinned;
pinning them works fine too :)

Issue #431, opened 2019-01-07
test don't terminate without `--exit`


tripodsan:
**Description**
the tests don't terminate without specifying `--exit`. this is a problem with the parcel-bundler workers not terminating properly.

**To Reproduce**

- run `npx mocha test/testUpCmd.js`
- the process doesn't terminate
- in a another terminal run `ps -ef | fgrep node` and see all the child processes

**Expected behavior**

the tests should terminate properly

**Version:**
`0.10.16-pre.2`

Issue #430, opened 2019-01-07
intermittent failures on CI due to node-watch


tripodsan:
**Description**
tests on CI sometimes file with:

```
Error: ENOENT: no such file or directory, stat '/home/circleci/repo/test/tmp/4a371c98-29bf-4894-903f-bc9d4b1d9f18/project/.git/objects/tmp_object_git2_4upySY'
```

this is due to a rise condition in:

https://github.com/yuanchuan/node-watch/blob/v0.6.0/lib/is.js#L40-L43

when a file is removed after `fs.exists` is invoked.

**To Reproduce**
intermittent.

**Expected behavior**
no error.

**Additional context**
this only happens on CI (so I guess a linux problem) and only below the `.git` directories. so a simple fix would be to exclude those form the file watcher.

tripodsan:
other possibility to replace `node-watch` with https://github.com/paulmillr/chokidar

PR #432, opened 2019-01-07
reduce test failures

tripodsan:
## Related Issues
see #431 and #430


tripodsan:
> you've been seeing high CPU usage under Linux,

I'll take a look also on linux. I think if the project is large, and the .git repository also, then the polling under linux might cause high cpu. the changes now only watch the `src` and the `helix-config.yaml`, instead of the entire _project directory_. this should also reduce the load.

koraa:
Are we using fs notification apis yet? There is the inotify api on linux which is quite fast, osx and windows probably have something similar…

tripodsan:
> Are we using fs notification apis yet? There is the inotify api on linux which is quite fast, osx and windows probably have something similar…

the used library https://github.com/paulmillr/chokidar relies on `fsevents`, which is only supported for osx and windows. on linux they do polling. I think the proper approach would be to add the inotify api support to fsevents.

PR #443, opened 2019-01-08
Better error handling

koraa:
Improves error messages

@trieloff Can you check whether the change in openwhisk/static.js makes sense?


src/openwhisk/static.js
```diff
@@ -129,13 +129,7 @@ function deliverPlain(owner, repo, ref, entry, root) {
...
       },
     };
-  }).catch((rqerror) => {
-    console.error('REQUEST FAILED', rqerror.response.body.toString());
-    if (rqerror.statusCode === 404) {
-      return error(rqerror.response.body.toString(), 404);
-    }
-    return error(rqerror.message, rqerror.statusCode);
-  });
+  }).catch(rqerror => error(rqerror.response.body.toString(), rqerror.statusCode));
```
trieloff: Always responding with `body.toString()` makes sense, but I'd differentiate between the actual status codes we are seeing from `raw.githubusercontent.com`:
- 404: is actually a 404
- 400 (when there is no file name, just a directory): should be reported as a 404
- other cases: just pass it through

trieloff:
I'd also make sure to hit 100% diff coverage, w^H I got into this error by not testing enough.

trieloff:
@koraa are you still working on this issue or should I take over?

koraa:
Sorry for the late response!

@trieloff I think it makes sense for you to take over the PR since you wrote the code and probably know better what needs testing and what does not :)

trieloff:
Will do.

Issue #413, opened 2018-12-20
deploy refused with unintuitive error message


rofe:
**Description**
When running `hlx deploy` on a fresh demo project, it will just say:
```
Invalid URL: undefined
```
**To Reproduce**
1. `hxl demo --type=full mytest`
2. `cd mytest`
3. 'hlx build`
4. `hlx deploy`

**Expected behavior**
`hlx deploy` should tell me that I need to either define a `code` repo in `helix-config.yaml`, or add a remote origin to my local git project.

**Version:**
0.10.12-pre.3

tripodsan:
will be fixed together with #231

Issue #231, opened 2018-10-04
Remove computed defaults from yargs options


tripodsan:
The defaults arguments for various commands, like the `contentRepo` of `hlx deploy` are indirectly set via the default values in yargs. this creates and implicit dependency from the commands to the yargs command definition. also, those defaults are evaluated during the initialization of yargs and not during command execution. for example the `--content`  default is read from the config file. so it will never be possible to specify a different config file via arguments.

## Suggestion
- Remove computed defaults from the yargs options.
- Compute them during the `validate()` step
- implement either a `hlx config` command that shows all defaults, or add a `--dump` to all commands that just dump their config and stop.

@trieloff @kptdobe  WDYT?

tripodsan:
also see #14

kptdobe:
I even did not know about the `--content` option. Thus yes, we should have a way to know all the possible arguments and their meaning.
Why not `hlx up --help` or something like a known pattern... ?

tripodsan:
> Why not hlx up --help or something like a known pattern... ?

`hlx up --help` already shows all the options :-).
but the problem is, that they are computed inside the yargs setup itself, instead provided by the command handlers.

but I just realized that `hlx build --help` is broken and doesn't show the arguments anymore...hmmm

tripodsan:
see #239

tripodsan:
@purplecabbage this fix probably eases the integration into the unified cli project.

PR #448, opened 2019-01-09
default prefix fixes

tripodsan:
fixes:
* Remove computed defaults from yargs options #231
* Use canonical action names independent of git origin #336
* [deploy] refuses to deploy with unintuitive error message if no remote git repo defined #413


tripodsan:
the coverage went down, because the files got shorter :-(
and `deploy.cmd` doesn't have good integration tests, as we don't actually deploy the actions in the tests.

PR #462, opened 2019-01-15
Fix helix-pipeline #157

koraa:

## Related Issues

Thanks for contributing!

### DIFF
--- a/demos/full/src/html.htl
+++ b/demos/full/src/html.htl
@@ -25,9 +25,7 @@
 </head>
 <body>
 
-<div data-sly-list="${content.children}">
-    <div class="container-fluid">${item}</div>
-</div>
+${content.html}
 
 <div class="container-fluid">
     <em>generated at ${content.time}</em>
--- a/demos/simple/src/html.htl
+++ b/demos/simple/src/html.htl
@@ -25,9 +25,7 @@
 </head>
 <body>
 
-<div data-sly-list="${content.children}">
-    <div class="container">${item}</div>
-</div>
+${content.html}
 
 <hr>
 <em>generated at ${content.time}</em>


demos/full/src/html.htl
```diff
@@ -25,9 +25,7 @@
 </head>
 <body>
 
-<div data-sly-list="${content.children}">
-    <div class="container-fluid">${item}</div>
-</div>
+${content.html}
```
rofe: not sure if it's good practice to eliminate the parent element for all "content" in the `<body>`... especially since according to the referenced https://github.com/adobe/helix-pipeline/issues/157, single elements aren't wrapped into a `<div>` element themselves...

i guess implementing projects will add one anyway, but the demo can serve as best practice here.

could you explain the the reason for removing it?

trieloff:
Fixes #157

demos/full/src/html.htl
trieloff: One thing that causes adobe/helix-pipeline#157 is that this code

https://github.com/adobe/helix-pipeline/blob/ed305629daf250ebfe339ca45ba539987902e7ad/src/html/emit-html.js#L16-L19

expects `document.body.firstChild.childNodes` to be truthy, which it isn't for simple text nodes.

Given that we want to deprecate `content.children` anyway (https://github.com/adobe/helix-pipeline/issues/105) it is only consequential to remove it from the examples.

However, this example HTL used to illustrate how to iterate over the content structure, which a simple `content.html` does not do, so a better fix would be to create an equivalent example that uses `data-sly-list="${content.document.body.children}"` and `{item.outerHTML}` (not sure if that would work, though)

tripodsan: @trieloff can we merge this in? the demos are currently broken otherwise.

trieloff: I'd still would like to show some iteration here, so let's try to incorporate the fix I've suggested above.

trieloff:
@rofe do you want to change your 👎 to a 👍?

demos/full/src/html.htl
rofe: Note that the current state of this branch will not work: `${content.html}` is undefined. It should be changed to `${content.document.body.innerHTML}` 

(even though I'd appreciate having a shortcut like `${content.html}` if that's what most projects will need... but that's something for a separate issue I guess)

rofe:
Btw, issue 157 doesn't seem right...
@trieloff @koraa

trieloff:
Closing as a duplicate of #573/#575

Issue #501, opened 2019-02-05
Incorrect directory path handling


kptdobe:
Check https://github.com/adobe/developer.adobe.com/issues/105 for all the details of the issue.

For https://alex.helix-demo.xyz/launch/docs, `req.url.dirname` equals `/launch`
For https://alex.helix-demo.xyz/launch/docs/README.html, `req.url.dirname` equals `/launch/docs`

Those 2 cases (basically without/out extension) must be treated separately.

PR #502, opened 2019-02-05
Incorrect directory path handling #501

kptdobe:
PR for  #501

### DIFF
--- a/src/fastly/vcl-utils.js
+++ b/src/fastly/vcl-utils.js
@@ -17,16 +17,10 @@ function conditions([strain, vcl]) {
     if (uri.path && uri.path !== '/') {
       const pathname = uri.path.replace(/\/$/, '');
       const body = vcl.body || [];
-      body.push('if (req.url.ext == "") {');
-      // use X-URL for root directory
-      body.push(`  set req.http.X-Dirname = regsub(req.http.X-URL, "^${pathname}", "");`);
-      body.push('} else {');
-      // use dirname for url with file and extensions
-      body.push(`  set req.http.X-Dirname = regsub(req.url.dirname, "^${pathname}", "");`);
-      body.push('}');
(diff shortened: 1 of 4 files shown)


kptdobe:
Still need to adjust the tests.

src/fastly/vcl-utils.js
```diff
@@ -17,10 +17,16 @@ function conditions([strain, vcl]) {
...
+      // use X-URL for root directory
+      body.push(`  set req.http.X-Dirname = regsub(req.http.X-URL, "^${pathname}", "");`);
+      body.push('} else {');
+      // use dirname for url with file and extensions
+      body.push(`  set req.http.X-Dirname = regsub(req.url.dirname, "^${pathname}", "");`);
+      body.push('}');
       return [strain, {
         sticky: false,
-        condition: `req.http.Host == "${uri.host}" && (req.url.dirname ~ "^${pathname}$" || req.url.dirname ~ "^${pathname}/")`,
+        condition: `req.http.Host == "${uri.host}" && (req.http.X-URL ~ "^${pathname}$" || req.http.X-URL ~ "^${pathname}/")`,
```
tripodsan: why testing against `X-URL` and not `X-Dirname` ?

kptdobe: because the condition leads to the creation of... `X-Dirname`. The code above will be in the if, see the test:

```
if (req.http.Host == "www.new-site.com" && (req.http.X-URL ~ "^/old-stuff$" || req.http.X-URL ~ "^/old-stuff/")) {
  if (req.url.ext == "") {
    set req.http.X-Dirname = regsub(req.http.X-URL, "^/old-stuff", "");
  } else {
    set req.http.X-Dirname = regsub(req.url.dirname, "^/old-stuff", "");
  }
```

But I am going to do it differently, I will introduce a new "global" variable that contains the full expected dir name and get rid of the if in here. Will be much more readable in the generated strains.vcl.

PR #508, opened 2019-02-06
OpenWhisk chokes on SVG if not base64 encoded

### LINKED ISSUE #507: OpenWhisk chokes on SVG if not base64 encoded
**Description**
This was found in https://github.com/adobe/developer.adobe.com/issues/88: Runtime/OpenWhisk will throw a 400 on SVG files spooled by the Helix static action. [Apparently](https://doc.akka.io/api/akka-http/10.0.4/akka/http/scaladsl/model/MediaTypes$.html#image/svg+xml:akka.http.scaladsl.model.MediaType.Binary), OpenWhisk will treat everything not explicitly containing `text` in its content-type as binary. Even though SVG is plain text in nature, it is expected to come base64 encoded.

**To Reproduce**
1. `hlx demo --type=full`
2. `hlx deploy`
3. Go to https://adobeioruntime.net/api/v1/web/<your_wsk_namespace>/default/hlx--static?owner=adobe&repo=developer.adobe.com&ref=master&strain=default&entry=/icons/spectrum-css-icons.svg&root=/htdocs&plain=true
4. See 400 error:
```
{
  "code": "<id>",
  "error": "Response type in header did not match generated content type."
}
```

**Expected behavior**
SVG files should be delivered with 200 and intact

**Version:**
0.13.3-pre.2

rofe:
Fixes #507 
Fixes https://github.com/adobe/developer.adobe.com/issues/88

### DIFF
--- a/src/openwhisk/static.js
+++ b/src/openwhisk/static.js
@@ -58,7 +58,7 @@ function isBinary(type) {
     return false;
   }
   if (type.match(/.*\/.*xml/)) {
-    return false;
+    return !!type.match(/svg/); // openwshisk treats svg as binary
   }
   return true;
 }
--- a/test/testStatic.js
+++ b/test/testStatic.js
@@ -34,11 +34,12 @@ describe('Static Delivery Action #unittest', () => {
     assert.equal(index.isBinary('application/octet-stream'), true);
     assert.equal(index.isBinary('image/png'), true);
     assert.equal(index.isBinary('un/known'), true);
+    assert.equal(index.isBinary('image/svg+xml'), true);
 
     assert.equal(index.isBinary('text/html'), false);
+    assert.equal(index.isBinary('text/xml'), false);
     assert.equal(index.isBinary('application/json'), false);
     assert.equal(index.isBinary('application/javascript'), false);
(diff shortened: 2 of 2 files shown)


src/openwhisk/static.js
```diff
@@ -58,7 +58,7 @@ function isBinary(type) {
     return false;
   }
   if (type.match(/.*\/.*xml/)) {
-    return false;
+    return !!type.match(/svg/); // openwshisk treats svg as binary
```
tripodsan: I think openwhisk treats all `image/*` as binary. maybe that would be the easier fix.

rofe: So if I understand you correctly, you mean adding a
```
if (type.match(/image\/.*/)) {
    return true;
}
```
above?

rofe: or rather wrap all the existing `if`s in a 
```
if (!type.match(/image\/.*/)) {
    ...
}
```

tripodsan: we should use the same code as openwhisk. I think it's here:
https://github.com/apache/incubator-openwhisk/blob/master/core/controller/src/main/scala/org/apache/openwhisk/core/controller/WebActions.scala#L325

but never mind for now.

PR #534, opened 2019-02-13
Static: emit events to notify Strains about changes

rofe:
Fixes: https://github.com/adobe/helix-shared/issues/49

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6524,21 +6524,6 @@
         }
       }
     },
-    "hasbin": {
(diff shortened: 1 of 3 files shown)


rofe:
> shouldn't the static-strain test fail?

No, the test never failed ;) we never tested the console output for `'localhost'`. I can add that though.

rofe:
Test: https://github.com/adobe/helix-cli/pull/548

PR #586, opened 2019-02-25
Force status code 200 for OK responses

### LINKED ISSUE #585: helix--static returns statusCode: 0 instead of 200
See 
https://github.com/adobe/developer.adobe.com/issues/133#issuecomment-466581959

trieloff:
fixes #585 might aid https://github.com/adobe/developer.adobe.com/issues/133#issuecomment-466768631


src/openwhisk/static.js
```diff
@@ -124,6 +124,7 @@ function deliverPlain(owner, repo, ref, entry, root) {
       const body = getBody(type, response.body);
       console.log(`delivering file ${cleanentry} type ${type} binary: ${isBinary(type)}`);
       return {
+        statusCode: 200,
```
tripodsan: how has this ever worked??

Issue #591, opened 2019-02-27
static files are sometimes not delivered via fastly


tripodsan:
see https://github.com/adobe/developer.adobe.com/issues/133

I think the main problem is, that the 404 from the `html` action is cached, and then the `vlc_hit` doesn't restart the request for invoking the `static` action.

I believe that the improvements of https://github.com/adobe/helix-publish/pull/7 would address this issue. in the mean time, a quick workaround is to ensure the `404` is not cacheable:

```
sub vlc_fetch {
...
  } elseif ((beresp.status == 404 || beresp.status == 204) && !req.http.X-Disable-Static) {
    # That was a miss. Let's try to restart.
    set beresp.http.X-Status = beresp.status + "-Restart " + req.restarts;
    set beresp.status = 404;
    # ensure that 404 from the backends are not cached. otherwise the a potential vlc_hit will not re-fetch the static files.
    set beresp.cacheable = false;    
...
```

PR #593, opened 2019-02-27
Fix static 404

tripodsan:
see #591 

also adds action name to trace for better debugging of the request chain.

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -681,6 +681,8 @@ sub vcl_fetch {
     # That was a miss. Let's try to restart.
     set beresp.http.X-Status = beresp.status + "-Restart " + req.restarts;
     set beresp.status = 404;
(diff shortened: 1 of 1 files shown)


tripodsan:
> Ideally, we _would_ cache the 404 coming from the static action, right?

yes, but not from the `html` action. I wouldn't invest too much here, as the refactored VCL works quite different. and, as I mentioned elsewhere, it would probably be better to only have 1 action that handles everything ;-)

ejthurgo:
It's common to cache 404s, but usually quite short lifespans compared to 2XX so perhaps explicitly specify a shorter TTL.
But I'm tempted to agree with Tobias, maybe worth waiting until the new VCL merges.

trieloff:
Ok, I'm merging this, but putting it on the https://github.com/adobe/helix-publish/pull/7 post-merge todo list.

PR #653, opened 2019-03-18
Inject `before` and `after` extension points

trieloff:
Support for adobe/helix-pipeline#211


test/integration/src/html.pre.js
```diff
@@ -21,3 +22,89 @@ module.exports.pre = (payload, action) => {
...
+    node.children[0] && 
+    node.children[0].children && 
+    node.children[0].children[0] && 
+    node.children[0].children[0].type === 'linkReference' &&
+    node.children[0].children[0].identifier && 
+    node.children[0].children[0].identifier.match(/![a-z]+/)) {
+    // get the class name
+    const classname = node.children[0].children[0].identifier.replace(/!/, 'admonition-');
+    const title = node.children[0].children[0].identifier.replace(/!/, '');
+
```
kptdobe: `node.children[0].children[0]` are scary (but only way to do it for now I think). In the future, we might need to add helper method on the mdast nodes to avoid this.

trieloff: I think with https://github.com/syntax-tree/unist-util-select I could have written it a bit more elegantly.

trieloff: At least that's what I'd have used if I had to write the same functionality for `helix-pipeline`

tripodsan: or use: https://lodash.com/docs/4.17.11#at

Issue #487, opened 2019-01-30
[hlx package] does not package "linked" dependencies


kptdobe:
**Description**
While developing, I `npm link` the helix-pipeline to the helix-cli. When I run `hlx deploy` on my project, the helix-pipeline is not deployed at all. The wsk action cannot be executed.

**To Reproduce**
1. In helix-pipeline: `npm install`, then `npm link`
1. In helix-cli: `npm install`, then `npm link`, then `npm link @adobe/helix-pipeline`
1. In a project, run `hlx build` and then `hlx package`

Open the `.hlx/build/html.info.json` and check the `externals` libraries: `@adobe/helix-pipeline` is not in the list.

**Expected behavior**
`@adobe/helix-pipeline` should be part of the `externals` libraries

**Version:**
run: `$ hlx --version`
0.12.1-pre.0

cc @tripodsan

kptdobe:
I identified the issue: https://github.com/adobe/helix-cli/blob/master/src/parcel/ExternalsCollector.js#L71

When linked, the module resource / path is not `node_modules/@adobe/helix-pipeline` but the path of the linked folder. The regex does not consider that case.

tripodsan:
a workaorund is to `npm pack` the dependency and the reference it via `file://`

eg:

```
$ cd helix-pipeline
$ npm pack
$ cd ../helix-cli
$ npm add ../helix-pipeline/helix-pipeline@0.1.2.tgz
...
```

PR #668, opened 2019-03-20
fix(packaging): Detected linked dependencies and include specially in…

tripodsan:
… action bundle

fix: #487

### DIFF
--- a/src/parcel/ExternalsCollector.js
+++ b/src/parcel/ExternalsCollector.js
@@ -60,6 +60,25 @@ class ExternalsCollector {
       devtool: false,
     });
 
+    // todo async
+    const pkgInfos = {};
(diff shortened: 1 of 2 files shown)


src/parcel/ExternalsCollector.js
```diff
@@ -69,9 +88,19 @@ class ExternalsCollector {
         stats.compilation.modules.forEach((mod) => {
           if (mod.resource) {
             const m = nodeModulesRegex.exec(mod.resource);
-            const modName = m ? m[2] : null;
+            let modName;
+            let modPath;
+            if (!m) {
+              // check if this is a linked package that is not inside a 'node_modules' directory
```
tripodsan: actually, I think we should always resolve by `package.json`....

Issue #401, opened 2018-12-18
Remote Publish/Remote Perf


trieloff:
I'd like to move most of the work that `hlx publish` does into a separate `helix-publish` OpenWhisk action/microservice. Same for `hlx perf` and `helix-perf`.

With this, we can massively reduce the complexity of the client-side applications (they only need to make one request to the micro service) and get a number of advantages:

- force feeding updates to VCL and other config files
- easier sharing of access tokens (e.g. for Calibre or Logging endpoints)

trieloff:
See adobe/helix-logging#5 and adobe/helix-cli#372

tripodsan:
why the extra indirection? would you create one for each customer (installed during `hlx-deploy`) or have a global one? how do you retrieve the fastly access tokens (the bot will be able to)?

in any case, we should keep the `publish` logic available as cli (and API). if not in `hlx` then `hlxctl`.

trieloff:
> why the extra indirection? 

For `publish`:

1. so that we can force-update the VCL and are not dependent on CLI updates
2. because we have significant parts happing remotely in `helix-logging` anyway (and can't move them away)

For `perf`:

- so that we can use a shared performance monitoring account and users don't need to bring their own Calibre credentials anymore

> would you create one for each customer (installed during hlx-deploy) or have a global one?

One global. See above.

PR #675, opened 2019-03-21
Enable remote publishing by default

### LINKED ISSUE #647: Allow specifying remote publish service URL
For `hlx publish --remote`, we should change `--remote` from a boolean flag to a string flag, with the string indicating the URL of the publishing microservice. This would allow us to do smoke tests of the `helix-publish` micro service.

trieloff:
(still can be disabled using `--no-remote`) (see #401).

Enable setting the remote publish service URL using `--api-publish` (fixes #647)

Thanks for contributing!

### DIFF
--- a/src/publish.cmd.js
+++ b/src/publish.cmd.js
@@ -154,7 +154,7 @@ class PublishCommand extends AbstractCommand {
     return this;
   }
 
-  withPublishAPI(value) {
+  withPublishAPI(_) {
     // does not do anything here
     return this;
   }
--- a/src/publish.cmd.js
+++ b/src/publish.cmd.js
@@ -154,6 +154,11 @@ class PublishCommand extends AbstractCommand {
     return this;
   }
 
+  withPublishAPI(value) {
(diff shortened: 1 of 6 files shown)


tripodsan:
+1 

- I don't like the duplicated code in here and in `helix-publish`. can you export an API from `helix-publish` so that it can be used here? I think that it is still valuable to publish something manually. e.g. for debugging.

- minor details rename to `RemotePublish`: https://github.com/adobe/helix-cli/blob/684ecc7bba5025611a53ef4e021bd0440b0b806f/src/publish.js#L68

trieloff:
@tripodsan my plan is to remove the local publish code entirely from `helix-cli`. If you want to publish/debug manually, you can deploy `helix-publish` yourself and use `--publish-api` with your URL.

tripodsan:
> my plan is to remove the local publish code entirely from helix-cli

hmm.. I don't like this so much. it's one more moving piece that might break and that is difficult to version control. but ok for now... ;-)

PR #676, opened 2019-03-21
Remote perf

trieloff:
This took longer than expected, but the PR replaces most of the `hlx perf` implementation. Instead of calling `calibre` directly, which requires each user to have their own set of Calibre credentials, it uses the `helix-perf` micro service (which authenticates against Fastly and uses shared Calibre credentials). This way, performance testing is limited to people (or bots) who can publish and the credential footprint of Helix is a bit reduced.

In addition, this PR brings up the coverage for the perf command to 100%, which is much easier now that the entire service is under our control.


tripodsan:
@trieloff do you still need this branch?

PR #688, opened 2019-03-25
Tests for the JSON pipeline and some documentation

kptdobe:
PR for https://github.com/adobe/helix-pipeline/issues/220

Using the pipeline without helix-cli does not really makes sense, so documented and tested the json rendering here.


kptdobe:
Arfff. how can the code coverage decrease since I only added a test and some documentation...

docs/usages.md
```diff
@@ -0,0 +1,65 @@
+# Usages
```
trieloff:
```suggestion
# Usage
```

I think this should be singular (not a native speaker either)

trieloff:
You removed six lines of code that used to be covered, so the relative coverage is down.

kptdobe:
> You removed six lines of code that used to be covered, so the relative coverage is down.

Well, no! I did not touch that file!!!

kptdobe:
ok, the test coverage check is not so smart... it compares to latest master which has been updated in the meantime. I merged master back, should be fixed now.

docs/usage.md
```diff
@@ -0,0 +1,65 @@
...
+One of the usage of the selectors (in combination with ESI - [Edge Server Includes](https://www.w3.org/TR/esi-lang)) could be to have multiple HTML rendering for your site. You could have 3 `htl` files (together with their `pre.js`):
+
+1. `html.htl`
+1. `homepage_html.htl`
+1. `others_html.htl`
+
+The `html.htl` file could look like this:
+
+```html
+<esi:include src="${payload.dispatch_url}" />
```
tripodsan: do we really want to promote this technique ?

docs/usage.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+The `htmp.pre.js` would be responsible to determine which of the `homepage_html.htl` or `others_html.htl` needs to be used for a given request. Here is an example where only the `index.md` file is rendered with the `homepage_html.htl` script:
+
+```js
+function pre(payload, action) {
```
tripodsan: please rename `payload` to `context`. so it matches the [schema docs](https://github.com/adobe/helix-pipeline/blob/master/docs/content.schema.md).

docs/usage.md
```diff
@@ -0,0 +1,65 @@
...
+  if (payload.request.path.match('index.html')) {
```
tripodsan: `match` is probably wrong here. it wold also match `index.html.xml`.

docs/usage.md
```diff
@@ -0,0 +1,65 @@
...
+
+module.exports.pre = pre;
+```
+
+### Custom extension
+
+If you to render a Markdown file using `xml` or `json` rendering, a default pipeline is provided for those extensions. You simply need to create a `<extension>.js` file which contains one `main` function export and that sets the `payload.response.body` content. Here is an example of a `json` renderer:
+
+```js
+module.exports.main = async function main(payload) {
```
tripodsan: `s/payload/context`

docs/usage.md
```diff
@@ -0,0 +1,65 @@
...
+  return {
+    response: {
+      body: payload.content
+    }
+  }
+};
```
tripodsan: pay attention to https://github.com/adobe/helix-pipeline/issues/223, in case we change this behaviour.

docs/usage.md
kptdobe: We do not have a better way for now to have 2 different renderings for one site.

tripodsan: true. somehow I don't like the extra roundtrip for the ESI include. ideally, the template selection logic, especially in this case when it is so simple, should happen in fastly....

kptdobe: yes, maybe but then it cannot be the selector. In any case, this is how it is today, thus documented as is. When we have something better, we update the doc!

docs/usage.md
kptdobe: Same, when we change the behaviour, we must update the doc!

PR #704, opened 2019-03-31
📝 update (concepts) docs

filmaj:

---

I made a few small changes and added a few links to try to make some things clearer, I tried to look at the project docs as a first-timer with no prior helix knowledge.

However, I'd like to use this PR as a conversation starting point for expanding the docs. I think adding some definitions for core concepts like strains, and clarifying what is meant by selector, needs to be done, but I could use some help with that. One more thing unclear from this document is how a static vs. a dynamic resource is treated by helix. I'll try to point these out as inline comments.


docs/concepts.md
```diff
@@ -1,37 +1,37 @@
...
-A request is received on the edge server (fastly) and is associated to a service configuration.
-The configuration contains a varnish cache script (VCL) that extracts processing information from
-the request. those are:
+An HTTP request is received on the edge server (fastly) and is associated to a [service configuration][fastly-services].
+The configuration contains a varnish cache script ([VCL][vcl]) that extracts processing information from
+the request. Those are:
 
 - requested resource
 - selector and extension
 - strain name
```
filmaj: We should define what a strain is, and link to that here.

docs/concepts.md
```diff
@@ -1,37 +1,37 @@
...
 
-A request is received on the edge server (fastly) and is associated to a service configuration.
-The configuration contains a varnish cache script (VCL) that extracts processing information from
-the request. those are:
+An HTTP request is received on the edge server (fastly) and is associated to a [service configuration][fastly-services].
+The configuration contains a varnish cache script ([VCL][vcl]) that extracts processing information from
+the request. Those are:
 
 - requested resource
 - selector and extension
```
filmaj: We should define what is meant by selector and extension here - I think extension means "file extension"?

docs/concepts.md
```diff
@@ -1,37 +1,37 @@
...
 - strain name
 
-based on this it looks up:
+Based on this it looks up:
 
 - backend type and url
 - content repository
 - adobeioruntime action or static server url
```
filmaj: As someone relatively new to helix, it wasn't clear to me that a static server was even a part of the equation until I read this line. Perhaps we need to explain the static server somewhere in these docs.

docs/concepts.md
rofe: selector is the URL part between the resource and the extension, e.g. /foo.**bar**.html

docs/concepts.md
rofe: I think the best (and only?) description we currently have is this:
https://github.com/adobe/helix-shared/blob/master/CONFIGURATION.md

filmaj: While this describes, sort of, the few properties strains contain (but it doesn't seem like an exhaustive list!), it doesn't set out a concise definition. That is what I'm looking for here. Any suggestions? Here is my attempt:

> A _strain_ is a configuration that maps HTTP request parameters to a content repository that is processed by one or more helix pipelines. Taken together, a helix configuration contains a set of strains that describe what content (optionally processed by what pipelines) is returned as responses to HTTP requests.

To me the above sounds sort of like a "router" in other kinds of web development.

Just random thoughts, but figured better to offer a suggestion than to simply point out the problem 😛

docs/concepts.md
filmaj: What purpose does the selector serve? I will use this as a start for a definitions section but I think we need more.

docs/concepts.md
rofe: Thanks for the proposal! I propose 2 minor changes and have 1 question:
- strain selection is not just HTTP request params, but also (even mainly?) URL based
- it should be mentioned that a strain is a combo of code, content and static repos
- should we mention GitHub, Fastly and Runtime, or abstract them away?
So something like:
> A _strain_ is the part of a Helix configuration that maps HTTP requests to a combination of GitHub repositories for code, content and static files. In the simplest case, a single repository can be used for all three. 
A Helix configuration consists of one or multiple strains, which result in the deployment of matching OpenWhisk actions to Adobe I/O Runtime, and instruct Fastly which of them to invoke in response to HTTP requests.

wdyt?

filmaj: I've added this to the PR, but I do think it is useful to somehow mention pipeline, no? A strain defines the main (or only parts of?) a helix pipeline, right?

docs/concepts.md
rofe: You can use selectors to render different variations of the same underlying resource. For example, you might choose to have an `/index.html` (default/empty selector), an `/index.toc.json` for a table of contents of the same content in JSON format, or an `/index.sitemap.xml` etc. developer.adobe.com is also using a bunch of selectors.

docs/concepts.md
rofe: Hmm, not really - I think a strain primarily defines where code and content should be taken from. Which pipeline gets triggered for which request is actually defined by the code, or the extensions being used... 

but I'll happily stand corrected corrected 😛

docs/concepts.md
rofe: This is likely outdated information: today as we know, static content gets proxied through the static action...

filmaj:
OK, I've rebased w/ latest master and added a sentence about how static resources are handled.

I don't feel I have a deep enough handle on the project to add more information or fill out the empty definitions. I will leave it up to one of the core members of the helix team to merge this in if they think it is a good contribution.

rofe:
Thanks @filmaj! I added something about resources and redirected the pipeline link to https://github.com/adobe/helix-pipeline/blob/master/README.md where its concept is described in detail.

If there are no objections, I think we can merge this now and continue working on it in an iterative fashion.

filmaj:
@rofe sounds good to me

PR #726, opened 2019-04-04
feat(env): better developer support through env variables

### LINKED ISSUE #14: Support for specifying developer options
See better options mentioned here: https://github.com/adobe/helix-cli/issues/14#issuecomment-457049862

---
<del>with more and more options to configure the various commands, it might make sense to support reading the options and flags from a config file.</del>

### <del>suggest:</del>

<del>use the `helix-config.yaml` and add the configuration options:</del>

<del>example</del>

```yaml
contentRepo: https://github.com/Adobe-Marketing-Cloud/reactor-user-docs

loglevel: info
build:
  - target: .hlx/build
  - files: src/*.htl
  - cache: true
  - minify: false

server:
  # ??? move this here?
  contentRepo: https://github.com/Adobe-Marketing-Cloud/reactor-user-docs
  port: 3000,
  browser: false
  
deploy:
  auto: false,
  apikey: ${WSK_AUTH}
  apihost: runtime.adobe.io
  namepsace: ${WSK_NAMESPACE}
  loghost: example.loggly.com
  logkey: ${HLX_LOGKEY}
  target: .hlx/build # maybe reuse build.target
  docker: nodejs8:latest
  default:
    flags: 'green'
```

tripodsan:
fix #14


test/fixtures/all.env
```diff
@@ -0,0 +1,57 @@
...
+HLX_LOG_LEVEL = debug
+HLX_LOG_FILE = test.log
+
+# deploy + package + clean + build + up
+HLX_TARGET = foo
+
+# build + up
+HLX_CACHE = true
+HLX_MINIFY = true
+HLX_FILES = *.htl,*.js
```
kptdobe: What happen if I change this config ? can I add more files ? Is this tested somewhere ?

kptdobe:
If this is the fix for #14 where you want to use the package.json `config` object, why the PR does not have extra `package.json` files to test if this works ?

tripodsan:
> If this is the fix for #14 where you want to use the package.json `config` object, why the PR does not have extra `package.json` files to test if this works ?

see the comment at the very bottom:
https://github.com/adobe/helix-cli/issues/14#issuecomment-472656272

we decided to not to do use the package.json approach, in favour of the `.env` files.

test/fixtures/all.env
tripodsan: > What happen if I change this config ? can I add more files ? Is this tested somewhere ?

well, it's the same as specifying the files in `hlx up` or `hlx build`. eg:

```
hlx up ./src/foo/*.js
```

it is implicitly tested via:
https://github.com/adobe/helix-cli/blob/v0.13.14/test/testBuildCmd.js#L41
and
https://github.com/adobe/helix-cli/blob/v0.13.14/test/testBuildCli.js#L93

PR #797, opened 2019-04-24
Use Static References in Static CSS and JS code

trieloff:
This is a partial implementation of adobe/helix-pipeline#224, depending on:
- https://github.com/adobe/helix-publish/pull/61 (to process `*.css.esi` requests)
- (optional) https://github.com/adobe/helix-pipeline/pull/281 (to create `*.css.esi` links in generated HTML)

For JS files:
- parses the file with babel
- replaces all `import foo from 'bar.js'` statements with an ESI

For CSS files:
- parses the file with postcss
- replaces all `@import url('bar.css');` statements with an ESI


tripodsan:
> replaces all `import foo from 'bar.js'` statements with an ESI

hmm, this is dangerous, I think. I wouldn't dare to rewrite JS...as most of the JS is anyways packaged and bundled and converted to non ES6 module style.

I would remove this for now.

trieloff:
The ESI is falling back to the original URI, and as we are using Babel for rewrites (the same tool that would be used by any bundler), we should be good.

tripodsan:
>The ESI is falling back to the original URI, and as we are using Babel for rewrites (the same tool that would be used by any bundler), we should be good. 

@trieloff but I don't think it is really useful. Since nobody used es6 imports in client js directly.

I'm just worried that this entire esi static will make things very unstable and unpredictable.

trieloff:
There is almost 90% support for ES6 in browsers today: https://caniuse.com/#feat=es6 and building fast sites without excessive bundling (in the spirit of https://www.pikapkg.com/blog/pika-web-a-future-without-webpack) is something we explicitly want to encourage.

The rewriting only happens for files that are requested to be static in the first place, so that
- https://www.primordialsoup.life/index.js.hlx_16116b678abb9c0c081f72320adc136f873169d1 will be rewritten
- but https://www.primordialsoup.life/index.js.hlx won't (but it won't be treated as immutable either)

PR #799, opened 2019-04-26
feat(publish): remove local publish capability

### LINKED ISSUE #795: remove non remote publish capability
with #468 and #675, `hlx publish` always uses the publish service. the VCL generation of the local and remote publish have diverged and should be unified again - or the local publish capabilities should be removed.

since we want to rely on the publish service, we should remove the local publish capability.

tripodsan:
BREAKING CHANGE: The local publish capability is removed and hlx publish completely relies on the
helix-publish service. the use of the --remote argument in hlx publish will now report an error.

fixes #795


tripodsan:
(note, there is little I can do against the drop in coverage...if you remove lines from a 90% covered file)

PR #803, opened 2019-04-29
test(testutils): use DOM instead of string comparison for integration tests

trieloff:


test/utils.js
```diff
@@ -100,6 +102,39 @@ async function assertHttp(url, status, spec, replacements = []) {
   });
 }
 
+async function assertHttpDom(url, status, spec) {
+  return new Promise((resolve, reject) => {
```
koraa: Didn't we have request-promise so we can avoid manual promisification?

test/utils.js
```diff
@@ -100,6 +102,39 @@ async function assertHttp(url, status, spec, replacements = []) {
...
+
+      res
+        .on('data', (chunk) => {
+          data += chunk;
+        })
+        .on('end', () => {
+          try {
+            if (spec) {
+              const datadom = new JSDOM(data).window.document;
+              const specdom = new JSDOM(fse.readFileSync(path.resolve(__dirname, 'specs', spec)).toString()).window.document;
```
koraa: Why not just store the html inline? There is `multiline()` to let that work nicely and if we are using promises we could just promise-load the file…

test/testUpCmd.js
```diff
@@ -211,7 +212,7 @@ describe('Integration test for up command', function suite() {
     });
 
     try {
-      await assertHttp(`http://localhost:${cmd.project.server.port}/README.html`, 200, 'simple_response_readme.html');
+      await assertHttpDom(`http://localhost:${cmd.project.server.port}/README.html`, 200, 'simple_response_readme.html');
```
koraa: The name seems a bit hard to understand…how about assertFetchedEquivalentDom

koraa: fetchAndAssertEquivNode…

trieloff:
@koraa answer to all your questions: I wanted to keep it in line with the other code in the file and wasn't in the mood for a big refactoring.

PR #814, opened 2019-05-02
fix(static): return the original URL in case a static resource cannot…

### LINKED ISSUE #813: URL of missing static resource replaced with 404 error
**Description**
The following HTL
```
<link rel="stylesheet" href="/i-do-not-exist.css"/>
```
will be turned into
```
<link rel="stylesheet" href='Error 404: 404: Not Found
'>
```
in production (including single quotes and line break).

**To Reproduce**
1. `hlx demo demo && cd demo`
2. Modify `src/html.htl` to reference an inexistent css file
3. `hlx` build, deploy, and publish the project
4. See error in source code

**Expected behavior**
Static URLs should remain as is, regardless of the HTTP status.

**Version:**
1.0.0

trieloff:
… get retieved for an ESI include

fixes #813

@rofe this is not yet entirely done, in case you want to take over while I'm working on VCL issues. I suggest you deploy this to see if it works as I hope and then write a unit test for the 404/esi=true case.

### DIFF
--- a/src/openwhisk/static.js
+++ b/src/openwhisk/static.js
@@ -44,6 +44,7 @@ function error(message, code = 500) {
     headers: {
       'Content-Type': 'text/html',
       'X-Static': 'Raw/Static',
+      'Cache-Control': 'max-age=300',
     },
     body: `${message}`,
   };
--- a/src/openwhisk/static.js
+++ b/src/openwhisk/static.js
@@ -266,7 +266,23 @@ function deliverPlain(owner, repo, ref, entry, root, esi = false) {
         'X-Static': 'Raw/Static',
       },
     };
-  }).catch(rqerror => error(rqerror.response.body.toString(), rqerror.statusCode));
+  }).catch(rqerror => {
+    if (esi) {
+      // the ESI failed, so we simply fall back to the original URL
+      // the browser will fetch it again, so let's cache the 404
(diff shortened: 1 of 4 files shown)


rofe:
@trieloff i just tried your branch, but the result is still the same. 

Debugging shows that the backend request to the static action still returns 
```
Error 404: 404: Not Found

```

rofe:
Looks like the expected `esi` param isn't included in the request to the static action...

rofe:
@trieloff can you check if the `esi` param gets set correctly in the VCL?

src/openwhisk/static.js
```diff
@@ -266,7 +267,26 @@ function deliverPlain(owner, repo, ref, entry, root, esi = false) {
...
+        statusCode: 404,
+        headers: {
+          'Content-Type': 'text/plain',
+          'Cache-Control': 's-max-age=300',
+        },
+        body: entry,
+      };
+    }
+    if (rqerror.statusCode === 404 || rqerror.statusCode === '404') {
+      return error(entry, rqerror.statusCode);
```
rofe: This looks like an attack vector for reflected XSS: `error()` will set a `text/html` response and use the `message` as `body`...

trieloff: done.

PR #834, opened 2019-05-07
Filter strains to publish with --only and --exclude

### LINKED ISSUE #821: How to enable Partial Publishing in a CI/CD workflow?
@filmaj    [Yesterday at 9:05 PM]
> i asked a few weeks back if a partial deploy / publish was possible, such that i could deploy+publish staging strains to one domain/host, run a battery of tests against that, and if that passes, then deploy+publish production strains, but that didnt seem possible

@trieloff    [10 hours ago]
> The problem is that the publish operation is atomic, i.e. you cannot partially update only a few strains, because the non-updated strains effectively disappear.

@trieloff    [10 hours ago]
> What you can do is to create staging strains that only work on a non-production domain, so no regular visitor can ever see them.

@trieloff    [10 hours ago]
> This still leaves you with a required edit operation once you promote it to master.

@trieloff    [10 hours ago]
> Maybe one way of building this (and I’d like to hear @dpfister’s take, too) is to add a CI or GitHub branch status as a possible condition to a strain. Here is how this could look like in a `helix-config.yaml`:
```strains:
  - name: subsite-staging
    condition:
      and:
        - url: https://stage.example.com/support/
  - name: subsite-production
    condition:
    and:
      - url: https://www.example.com/support/
      - ci: https://circleci.com/gh/adobe/helix-publish/tree/fresh-config
```

> The `subsite-staging` strain can be entered simply by going to `stage.example.com`, but the `subsite-production` strain is only enabled when CircleCI reports tests as passing on https://circleci.com/gh/adobe/helix-publish/tree/fresh-config. Wether the test is passing or not should be evaluated at `publish` time, so that it does not flip flop randomly.
You could then have a development flow where the developer who’s working on a new feature (@hireshah) creates his branch, with the two new strains and submits a PR. The maintainer (@maj) reviews it and merges it even when the branch is still not completely ready, because he has confidence that no production traffic will meet the strain until the CI build is green. (edited)

@dominique-pfister    [8 hours ago]
> Hm, interesting :slightly_smiling_face: But then the decision must be somehow persisted, right?

@trieloff    [8 hours ago]
> We could store this in an edge dictionary, simply using the CircleCI or GitHub URL as a key, or we could just store it in the plain VCL file.

@dominique-pfister    [8 hours ago]
> If the VCL is generated on publish anyway, that would be straightforward, of course. (edited)

@filmaj    [1 hour ago]
> It’s an interesting idea but but tests-passing is sometimes a false positive as well (“i forgot to add tests to my new feature”).
what about having two helix configs, one for staging and one for production? then continuous deployment would look like:
```hlx deploy --config helix-config-staging.yaml && npm run runtime-tests-staging && hlx publish --config helix-config-staging.yaml && npm run end-to-end-tests-staging && hlx deploy --config helix-config-production.yaml && npm run runtime-tests-production && hlx publish --config helix-config-production.yaml && npm run end-to-end-tests-production && npm run tag-known-good```

@filmaj    [38 minutes ago]
> is there a better venue for this discussion than a slack thread? this is quite urgent from our end. we’ve hit a few issues now in the transition from local `hlx up` dev to deployed-to-fastly, so we need to nail down a way to publish to, verify at and promote from a publicly accessible staging env to production

@filmaj    [11 minutes ago]
> i appreciate your suggestion’s use of automation, and its a cool idea, but i would also like the ability to promote and demote manually, especially as we build out the full test suite and figure it all out

@trieloff    [6 minutes ago]
> With the new conditions language, you will be able to easily promote and demote strains, because you can simply mix in a `*prod` or `*stage` into a condition.

@trieloff    [4 minutes ago]
> I don’t really like the idea of separate prod and stage configurations because it discourages incremental, feature-flag driven development and feels too much like having environments, which aren’t really a thing in serverless. (or shouldn’t be)

### PR DESCRIPTION
**Pull request by trieloff**
Fixes #821 (work in progress)

### DIFF
--- a/src/publish.js
+++ b/src/publish.js
@@ -59,7 +59,7 @@ module.exports = function strain() {
           type: 'string',
         })
         .option('exclude', {
-          describe: 'Do publish strains with names following the specified pattern, use config from master branch instead',
+          describe: 'Don\'t publish strains with names following the specified pattern, use config from master branch instead',
           type: 'string',
         })
         .conflicts('only', 'exclude')
--- a/src/publish.js
+++ b/src/publish.js
@@ -54,6 +54,15 @@ module.exports = function strain() {
           describe: 'Update the helix bot configuration on the affected content repositories.',
           type: 'boolean',
         })
+        .option('only', {
+          describe: 'Only publish strains with names following the specified pattern, use config from master branch for all others',
+          type: 'string',
+        })
+        .option('exclude', {
+          describe: 'Do publish strains with names following the specified pattern, use config from master branch instead',
+          type: 'string',
+        })
+        .conflicts('only', 'exclude')
         .demandOption(
           'fastly-auth',
           'Authentication is required. You can pass the key via the HLX_FASTLY_AUTH environment variable, too',
@@ -96,6 +105,7 @@ module.exports = function strain() {
         .withGithubToken(argv.githubToken)
         .withUpdateBotConfig(argv.updateBotConfig)
         .withConfigPurgeAPI(argv.apiConfigPurge)
+        .withFilter(argv.only, argv.exclude)
         .run();
     },
   };
--- a/package-lock.json
+++ b/package-lock.json
@@ -11054,7 +11054,7 @@
       "dependencies": {
         "JSONStream": {
           "version": "1.3.4",
-          "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.4.tgz",
+          "resolved": false,
           "integrity": "sha512-Y7vfi3I5oMOYIr+WxV8NZxDSwcbNgzdKYsTNInmycOq9bUYwGg9ryu57Wg5NLmCjqdFPNUmpMBo3kSJN9tCbXg==",
           "dev": true,
           "requires": {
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: src/publish.js [hunk 1]
```diff
@@ -54,6 +54,15 @@ module.exports = function strain() {
           describe: 'Update the helix bot configuration on the affected content repositories.',
           type: 'boolean',
         })
+        .option('only', {
+          describe: 'Only publish strains with names following the specified pattern, use config from master branch for all others',
+          type: 'string',
+        })
+        .option('exclude', {
+          describe: 'Do publish strains with names following the specified pattern, use config from master branch instead',
```
tripodsan:
```suggestion
          describe: 'Don\'t publish strains with names following the specified pattern, use config from master branch instead',
```

trieloff:
@kptdobe 

> What about the "partial" deploy part ? If I create a new strain, I need to `hlx deploy mynewstrain` first and then `hlx publish mynewstrain`.

My assumption is that this would be used in a CI workflow. With branch-specific configurations. On `master` you `hlx deploy` everything and `hlx publish` everything. On any other branch, you would `hlx deploy` everything (but that won't affect production, as deployments are SHA-tagged) and `hlx publish --exclude "prod-*"` or `hlx publish --only "stage-*" or `hlx publish --only "stage-$BRANCHNAME"`, so that only the strains that may be affected by the current branch will see changes.

> To validate the use case described in #821, I would first create multiple test scenarii and make sure the changes allow to do so. Maybe there is something I do not see but I think this address only a little part of the problem and in the end does not answer to the need. For me, going from *all-or-nothing* to *selective publication* has a lot implications and I have hard time to believe that those few lines of codes are enough.

My hope is that we can build a relatively small and lightweight feature (selective publishing) that will give @filmaj a tool to build and validate a more robust publishing flow. If it turns out that other features are needed, we will build them. If it turns out that this is not useful, we will remove it again. 

I certainly do not have the vantage point right now to draw a master plan, so I'd rather see what we can do to improve the status quo through incremental changes until I do, and then we can rethink the entire configuration/deployment/publishing complex.

kptdobe:
> My hope is that we can build a relatively small and lightweight feature (selective publishing) that will give @filmaj a tool to build and validate a more robust publishing flow. If it turns out that other features are needed, we will build them. If it turns out that this is not useful, we will remove it again.

ok, that's the way to go ;)

filmaj:
> My hope is that we can build a relatively small and lightweight feature (selective publishing) that will give @filmaj a tool to build and validate a more robust publishing flow. If it turns out that other features are needed, we will build them.

I'm on board with this plan!

Issue #203, opened 2018-09-18
Provide an example using JSX


trieloff:
I think we have built-in support for JSX, we just don't know it yet.

```jsx
/** @jsx h */

import h from 'hyperscript';

const App = (logo) => (
  <div className="App">
    <img className="App-Logo" src={logo} alt="React Logo" />
    <h1 className="App-Title">Hello Parcel x React</h1>
  </div>
);

console.log(App("test.png").outerHTML);
```

(https://github.com/hyperhype/hyperscript has a `React.createElement`-compatible (enough) API and the `/** @jsx h */` line instructs Babel to use `h` instead of `React.createElement`)

and

```bash
$ npx parcel build -t node src/*.jsx
$ node dist/html_test.js
<div class="App"><img class="App-Logo" src="test.png" alt="React Logo"><h1 class="App-Title">Hello Parcel x React</h1></div>
```

All that needs to be done is to create a proper `main` function that accepts the pipeline payload and returns the correct HTML.

tripodsan:
don't you think it will be too confusing for people to mix a templating language that is primarily used clientside for serverside?

trieloff:
No, there are a number of projects (e.g. next.js) that use JSX (and React) for server-side rendering.

dplaton:
Would we use JSX to provide an alternative to HTL?

tripodsan:
we could, yes. but I still think we shouldn't :-)

trieloff:
In principle, Helix is agnostic to your choice of templating language and JSX has a couple of advantages over HTL:

- it allows easy mixing of markup and code (so you often don't need a pre.js)
- it is widely known

In order to demonstrate Helix's openness, I think it would well serve it's purpose.

PR #855, opened 2019-05-09
feat(build): add basic JSX support

trieloff:
so far only *html.jsx is supported and no examples or tests are included

implements #203

Using the code from here: https://github.com/trieloff/helix-demo/blob/test-branch/src/foo_html.jsx you can create output like this:

```javascript
{ statusCode: 200,
  headers:
   { 'Content-Type': 'text/html',
     'Cache-Control': 's-maxage=604800',
     'Surrogate-Key': 'AYeGDnUcZX/lT0R/omdfQTuQ6J7P9D2fSTpvpzUFacs=' },
  body:
   '<div class="App"><img class="App-Logo" src="foo.jpg" alt="{title}"><h1 class="App-Title">Helix - demo</h1></div>',
  error: undefined }
```

### DIFF
--- a/src/parcel/OutputTemplate.js
+++ b/src/parcel/OutputTemplate.js
@@ -29,7 +29,6 @@ function helix_wrap_action(main) {
     async function once(payload, action) {
       // calls the pre function and then the script's main.
       async function invoker(next) {
-        console.log('invoking', next);
         const ret = await Promise.resolve(pre(payload, action));
         return Promise.resolve(next(ret || payload, action));
       }
--- a/src/parcel/OutputTemplate.js
+++ b/src/parcel/OutputTemplate.js
@@ -12,7 +12,7 @@
 
 /* eslint-disable */
 
-const h = require('hyperscript');
+const h = require('hyperscript'); // lgtm [js/unused-local-variable]
 
 
 // CONTENTS
--- a/src/parcel/OutputTemplate.js
+++ b/src/parcel/OutputTemplate.js
@@ -12,11 +12,7 @@
 
 /* eslint-disable */
 
-const h = require('hyperscript'); // lgtm [js/unused-local-variable]
-
-
 // CONTENTS
-
 
 function helix_wrap_action(main) {
   const { OpenWhiskAction } = require('@adobe/helix-pipeline');
--- a/src/parcel/OutputTemplate.js
+++ b/src/parcel/OutputTemplate.js
@@ -12,7 +12,11 @@
 
 /* eslint-disable */
 
+const h = require('hyperscript');
+
+
 // CONTENTS
+
 
 function helix_wrap_action(main) {
   const { OpenWhiskAction } = require('@adobe/helix-pipeline');
@@ -25,6 +29,7 @@ function helix_wrap_action(main) {
     async function once(payload, action) {
       // calls the pre function and then the script's main.
       async function invoker(next) {
+        console.log('invoking', next);
         const ret = await Promise.resolve(pre(payload, action));
         return Promise.resolve(next(ret || payload, action));
       }
--- a/README.md
+++ b/README.md
@@ -464,13 +464,83 @@ Please note that these languages are all executed server-side (or serverless-sid
 
 HTL stands for [HTML Template Language and was originally introduced for Adobe Experience Manager](https://docs.adobe.com/content/help/en/experience-manager-htl/using/getting-started/update.html). The implementation in Helix is based on the [HTL Specification](https://github.com/adobe/htl-spec/blob/master/SPECIFICATION.md), but as Helix and the underlying [`htlengine`](https://github.com/adobe/htlengine) are written in JavaScript rather than Java, and as the object model between Helix and AEM is different (check out the [`helix-pipeline` documentation](https://github.com/adobe/helix-pipeline/tree/master/docs) for Helix' domain model), your templates translate roughly rather than directly.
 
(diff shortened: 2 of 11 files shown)


tripodsan:
hmmm, beside the fact, that I don't think we should support jsx on the server side...,
shouldn't it just be:

**foo_html.jsx`
```
  <div className="App">
    <img className="App-Logo" src="foo.jpg" alt="{payload.content.title}" />
    <h1 className="App-Title">{payload.content.title}</h1>
  </div>
```

to be symmetrical to HTL?

trieloff:
JSX does not consider itself a template language like HTL, but a JavaScript syntax extension. All JSX examples I've found use a plain JS entry point such as: https://reactjs.org/docs/introducing-jsx.html#embedding-expressions-in-jsx

```jsx
const name = 'Josh Perez';
const element = <h1>Hello, {name}</h1>;

ReactDOM.render(
  element,
  document.getElementById('root')
);
```

The `ReactDOM.render` extension point clearly makes no sense for us, so I've just kept it with a normal pipeline boilerplate.

src/parcel/OutputTemplate.js
```diff
@@ -12,8 +12,12 @@
 
 /* eslint-disable */
 
+const h = require('hyperscript'); // lgtm [js/unused-local-variable]
```
tripodsan: hmm, I don't like that we have this dependency also for HTL...
can't we create a separate `OutputTemplate.js`  for JSX ?

src/parcel/OutputTemplate.js
```diff
@@ -12,8 +12,12 @@
...
+
+
 // CONTENTS
 
+
```
tripodsan: extra newline?

src/parcel/OutputTemplate.js
trieloff: We could, but the dependency is already in `helix-pipeline`.

trieloff:
I think we get source maps out of the box through Babel

src/parcel/OutputTemplate.js
tripodsan: but where is `h` used?

trieloff: I forgot that so far its only used in the tests. How would you pass through the output template selection?

tripodsan: > I forgot that so far its only used in the tests

I don't see it :-)

tripodsan: Ah, I see...it's created by parcel or babel.

tripodsan: it would be better, if we could define the `h` in the script and not in the output template. I'll take a look if this is possible.

tripodsan:
what about like this:

```diff
diff --git a/src/parcel/AdapterJSAsset.js b/src/parcel/AdapterJSAsset.js
index 8f009e0..305de16 100644
--- a/src/parcel/AdapterJSAsset.js
+++ b/src/parcel/AdapterJSAsset.js
@@ -61,6 +61,10 @@ class AdapterJSAsset extends JSAsset {
       return gen;
     }

+    // inject hyperscript library
+    if (this.basename.endsWith('.jsx')) {
+      gen[0].value = `const h = require('hyperscript');\n${gen[0].value}`;
+    }
     // otherwise it's a pure-JS action and we want to wrap it
     gen[0].type = 'helix-js';
     return gen;
diff --git a/src/parcel/OutputTemplate.js b/src/parcel/OutputTemplate.js
index 356bb32..7022197 100644
--- a/src/parcel/OutputTemplate.js
+++ b/src/parcel/OutputTemplate.js
@@ -12,12 +12,8 @@

 /* eslint-disable */

-const h = require('hyperscript'); // lgtm [js/unused-local-variable]
-
-
 // CONTENTS

-
 function helix_wrap_action(main) {
   const { OpenWhiskAction } = require('@adobe/helix-pipeline');
   const { pipe } = require('MOD_PIPE');
```

src/parcel/OutputTemplate.js
trieloff: The JSX support is pretty baked in in babel, and the examples I've seen use the annotation
```
/** @pragma h */
const h = import('hyperscript');
```

that (two-pass processing with Babel) could be an option for us.

tripodsan: I find the PRAGMA computation a bit hairy in parcel:
https://github.com/parcel-bundler/parcel/blob/master/packages/core/parcel-bundler/src/transforms/babel/jsx.js

 I guess, this might cause problems if someone has a different engine in their `package.json`....

trieloff:
@tripodsan I'd pseudo-randomly generate the name for `h` (maybe `h` + hash(file)) to avoid conflicts and then remove the dependency-yoga I'm doing to trick Babel into thinking it's based on hyperapp (which also uses the `h` prefix)  and instead add the annotation comment.

tripodsan:
> I'm doing to trick Babel into thinking it's based on hyperapp

I don't understand. 

the 1 pass is done by parcel/babel and generates:

```
const LinkList = (title, document) => h("footer", {
  className: "link-list"
}, h("h1", null, document.querySelectorAll("a[href]").length, " Links for ", title), h("ul", null, [...document.querySelectorAll("a[href]")].map(link => h("li", null, link))));
```

the second pass is done by the `OutputTemplate`...I just move the `const h = require` to our asset processing. At the time the `AdapterJSAsset.js` gets the source, it is already transformed by Babel.

trieloff:
Ok, let's try this. Please commit it straight to the branch.

src/parcel/OutputTemplate.js
trieloff: > I guess, this might cause problems if someone has a different engine in their `package.json`…

Let's burn that bridge when we get to it.

PR #862, opened 2019-05-10
feat(script): sanitize context before returning it to the pipeline

### LINKED ISSUE #744: not possible to set response headers in `pre.js`
**Description**
trying to set response headers in `pre.js` doesn't work, because it always overwrites the `response` object inside the `context. altering other properties of the `context` works.

the merging / response handling of pipeline steps is discussed here:
https://github.com/adobe/helix-pipeline/issues/223

**Additional context**
It is not yet clear if this is the intended behaviour, as the `pre.js` should probably not alter response headers directly. 

in any case, the proper approach is to use an extension:

```
// html.pre.js
...
module.exports.after = {
  html: (context) => {
    return {
      response: {
        headers: {
          'Cache-Control': 'private',
        }
      }
    };
  }
};
```

(we should probably expose the `cache` step as extension, or even better have an `emit` step.

### LINKED ISSUE #861: Adjust OutputTemplate to cope with new pipeline semantics
https://github.com/adobe/helix-pipeline/pull/260 will alter the pipeline semantics.
the major change is that the return value of a pipeline step is no longer merged into the `context`.

the `OutputTemplate` needs to be adjusted to copy with this change.
this can be done already, since the current pipeline already allows for modification of the `context`

tripodsan:
fixes #744
fixes #861

### DIFF
--- a/src/parcel/OutputTemplate.js
+++ b/src/parcel/OutputTemplate.js
@@ -19,14 +19,46 @@ function helix_wrap_action(main) {
   const { pipe } = require('MOD_PIPE');
   const { pre, before, after } = require('MOD_PRE');
 
+  // todo: mode to helix-pipeline
+  const CONTEXT_PROPS = ['error', 'request', 'content', 'response'];
+  const CONTENT_PROPS = ['sources', 'body', 'mdast', 'sections', 'document', 'htast' ,'json', 'xml', 'meta', 'title', 'intro', 'image'];
+  const REQUEST_PROPS =  ['url', 'path', 'pathInfo', 'rootPath', 'selector', 'extension', 'method', 'headers', 'params'];
+  const RESPONSE_PROPS = ['status', 'body', 'hast', 'headers'];
+
+  const filterObject = (obj, allowedProperties) => {
+    if (!obj) {
+      return;
+    }
+    Object.keys(obj).forEach((key) => {
+      if (allowedProperties.indexOf(key) < 0) {
+        delete obj[key];
+      }
+    })
+  };
+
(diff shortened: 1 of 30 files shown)


src/parcel/OutputTemplate.js
```diff
@@ -19,14 +19,46 @@ function helix_wrap_action(main) {
...
-        return Promise.resolve(next(ret || payload, action));
+        const ret = await Promise.resolve(pre(context, action));
+        const res = await Promise.resolve(next(ret || context, action));
+        if (res && res.response && res.response.body) {
+          if (!context.response) {
+            context.response = {};
+          }
+          context.response.body = res.response.body;
+        }
+        sanitizeContext(context);
```
trieloff: I'm not sure why you are doing this here. If we don't filter the context, the validator in the pipeline would complain anyway, so it can't be for protection against invalid properties?

tripodsan: the problem is the following:

for feeding the HTL with data provided by the `pre.js`, they need to add properties to `context`. (this will become the `payload` object in HTL...). until now, we never returned the modified `context` from the `pre.js` back to the pipeline, only the result of the HTL.
since the `cont()` had a return value, it was merged with the `context` that was present before the `cont()`.

now, if we abandon the merge behaviour, we need to keep the `context` that was modified in the `pre.js`. as mentioned in the issue, the proper solution would be to offer a `context.data` property, that could be used for anything. 

but for now, we just remove all _garbage_ from the context, before continuing the pipeline. otherwise the validator would complain.

PR #882, opened 2019-05-14
fix(parcel): more robust handling of missing package.json

### LINKED ISSUE #881: hlx up: cannot read property 'devDependencies' of undefined
**Description**
I'm getting a `Cannot read property 'devDependencies' of undefined` after running hlx demo and hlx up from a empty folder.

**To Reproduce**

    export H=...full path of helix-cli/index.js
    mkdir empty-folder
    cd empty-folder
    $H demo xxx
    cd xxx
    $H up

which says:

    ....html.pre.js: Cannot read property 'devDependencies' of undefined
    at AdapterJSAsset.getPackage (...helix-cli/src/parcel/AdapterJSAsset.js:43:9)

**Version:**
at commit 0855c48

trieloff:
fixes #881

@bdelacretaz does this fix your issue?

### DIFF
--- a/src/parcel/AdapterJSAsset.js
+++ b/src/parcel/AdapterJSAsset.js
@@ -39,7 +39,7 @@ class AdapterJSAsset extends JSAsset {
   }
 
   async getPackage() {
-    const pkg = await super.getPackage();
+    const pkg = await super.getPackage() || {};
     const pack = pkg.devDependencies ? pkg : Object.assign(pkg, {
       devDependencies: {
         hyperapp: '*',
--- a/src/parcel/AdapterJSAsset.js
+++ b/src/parcel/AdapterJSAsset.js
@@ -67,6 +67,11 @@ class AdapterJSAsset extends JSAsset {
       }
     }
   }
+
+  async getPackage() {
+    const pack = await super.getPackage();
+    return pack || {}; 
+  }
 }
 
 module.exports = AdapterJSAsset;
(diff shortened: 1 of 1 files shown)


bdelacretaz:
It does fix it but it's not setting `pkg.devDependencies` anymore, dunno if that's required. 

I was just submitting my own fix, #883, I'll let you select the right one!

trieloff:
I just noticed. Both a missing `package.json` and missing `devDependencies` would be an issue.

bdelacretaz:
I think you need

    const pkg = await super.getPackage() || {}

or something like that in the current fix, as pkg itself might be null. It's still failing for me with that fix.

My fix at https://github.com/adobe/helix-cli/pull/883/files works (but maybe it's not the preferred style)

src/parcel/AdapterJSAsset.js
```diff
@@ -39,9 +39,14 @@ class AdapterJSAsset extends JSAsset {
   }
 
   async getPackage() {
-    const pkg = await super.getPackage();
-    pkg.devDependencies.hyperapp = '*';
-    return pkg;
+    const pkg = await super.getPackage() || {};
+    const pack = pkg.devDependencies ? pkg : Object.assign(pkg, {
+      devDependencies: {
+        hyperapp: '*',
```
tripodsan: this is set twice now...

trieloff: Only when `devDependencies` didn't exist before.

tripodsan: why do you need it, btw ?

trieloff: In order to get `parcel` to think that the pragma for JSX files is `h` instead of `React.createElement` without forcing people to add it to their JSX files.

kptdobe:
I do not see a newly added test that would prevent for this bug to happen again...

PR #883, opened 2019-05-14
#881: Guard against missing pkg or devDependencies

### LINKED ISSUE #881: hlx up: cannot read property 'devDependencies' of undefined

bdelacretaz:
This fixes #881 - not sure if that's the preferred way of fixing it but it works for me.

### DIFF
--- a/src/parcel/AdapterJSAsset.js
+++ b/src/parcel/AdapterJSAsset.js
@@ -39,7 +39,10 @@ class AdapterJSAsset extends JSAsset {
   }
 
   async getPackage() {
-    const pkg = await super.getPackage();
+    const pkg = await super.getPackage() || {};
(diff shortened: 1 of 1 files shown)


bdelacretaz:
#882 has fixed this, thanks!

PR #886, opened 2019-05-14
Optionally bind static action instead of building it

### LINKED ISSUE #850: Move helix-static to separate repo, then bind it instead of deploying it
I'd suggest to change the deployment flow for `helix--static` from:

1. customer builds `helix--static` from current helix-cli version (risk of using an old version)
2. customer deploys `helix--static` into their own Runtime namespace (takes time, especially on @filmaj's home network 😉)

to:

1. `helix-static` lives in its own repository
2. CircleCI deploys `static` into the `helix` namespace, under the package name `helix-shared`, i.e. `/helix/helix-shared/static`
3. customers [bind](https://github.com/apache/incubator-openwhisk/blob/master/docs/packages.md#creating-and-using-package-bindings) (i.e. mount) the `/helix/helix-shared` package into their own namespace, so that they can run it against their own rate limits and provide it with their own credentials, but the code lifecycle is determined by us

trieloff:
Fixes #850, still WIP. Missing:

- [x] implement binding of static action
- [x] make building/binding conditional
- [x] tell `helix-publish` which static option has been chosen

This way we can gradually move to binding over building.

### DIFF
--- a/src/yargs-static.js
+++ b/src/yargs-static.js
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2019 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+module.exports = function commonArgs(yargs) {
+  return yargs
+    .option('static', {
+      describe: 'How to create the static action.',
+      type: 'string',
+      choices: ['both', 'bind', 'build'],
+      default: 'both',
+    });
+};
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "check": "npm run lint && npm run test",
     "test": "nyc --reporter=text --reporter=lcov mocha -t 5000",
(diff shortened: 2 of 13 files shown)


src/yargs-static.js
```diff
@@ -0,0 +1,21 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+module.exports = function commonArgs(yargs) {
+  return yargs
+    .option('static', {
+      describe: 'How to create the static action.',
+      type: 'string',
+      choices: ['both', 'bind', 'build'],
```
tripodsan: - what's the benefit of `both` ?

- why not using an array?
```
hlx deploy --static=bind --static=build
```

trieloff: I expect the `deploy` part of the static refactoring to land before the `publish` part. With `both`, you can switch more easily, because you will have both options deployed way before you start using the new publish feature.

tripodsan: ?? the linking should also be done during the `deploy` phase. to be consistent with the rest of the actions.

trieloff: When you use `both`, you will have two resources available:

1. `/helix--static`
2. `/helix-services/static@v1`

which one to use depends on the VCL `helix-publish` creates. So the actual choice is happening in the `publish` step.

trieloff: > the *linking* should also be done during the `deploy` phase. to be consistent with the rest of the actions.

*binding*, not linking, yes.

tripodsan: > When you use `both`, you will have two resources available:

I thought we should get rid of the private statics....

> which one to use depends on the VCL `helix-publish` creates. So the actual choice is happening in the `publish` step.

if we would put all the actions (even the private ones, if we decide to keep it) below `$actionpackage/helix--static`, the VLC could use the same logic for both cases.

> _binding_, not linking, yes.

I think both should be done during `deploy`:
- _binding_ (ensure that the helix-shared package is available in the customers namespace)
- _link_ (create the sequence action at `$actionpackage/helix--static`, analogous to create the normal script actions.

Issue #812, opened 2019-05-02
As an advanced developer, I would like to extend some parts of the VCL logic


kptdobe:
@koraa and myself worked both on a way to use the subdomain as the source for the content repo and owner. The use case is simple: 
- as a domain owner, I want to offer anyone the capability to hit https://<owner>-<repo>.mydomain.com:
  - a design (`html.htl` and `html.pre.js`) is handled by the domain owner (standard Helix project)
  - <owner> / <repo> can be defined by any end-user and refer to a git repo with some content.

To achieve that, it is pretty easy to setup an Helix project, deploy and publish. The handling of the <owner> / <repo> from the subdomain must be done on Fastly in VCL - basically requires a custom `helix.vcl` which gets overwritten each time by `hlx publish` thus requires a manual task to patch it in Fastly after each publication. Here is an example of customisation: https://gist.github.com/kptdobe/bd0f5bb10b88783a3cefd917beb98a68

The use case could be achieved in a very simple way if as an advanced developer I have a way to define a `custom.vcl` file in my Helix project that would be uploaded for me during the publish and that would allow me to extend / override certain pieces of the VCL. Like this, no need to manually patch the `helix.vcl` file.

For this use case, the extension point could be: a `hlx_type_pipeline_ext` subrouting is called at the end of the `hlx_type_pipeline` subroutine. By default it is empty, I would write one that contains the logic to override `req.http.X-Owner` and `req.http.X-Repo` based on the use case requirements.

Note 1:  this requires advanced VCL knowledges thus it should be reserved to really advanced developers that knows what they are doing.

Note 2: this is the only known extension for now, I would not allow more until we have a concrete and clear use case.

trieloff:
I think you will need the ability to insert before and after, as some subroutines use `return`.

kptdobe:
I started to look into since but I do not feel comfortable doing that. Before changing anything, I'd like to make sure I am taking the right approach. I would do it like this:

`helix-publish` side: 
- in `helix.vcl`, add a call to `hlx_type_pipeline_before` and `hlx_type_pipeline_after` at the end of the  `hlx_type_pipeline` subroutine. We do not need more for now.
- extend the `helix-publish` action to accept a new parameter: `customVCL` - string that my contain the VCL extensions.
  - a better approach would be to extend the `configuration` parameter but it seems that it is 1:1 the helix-config... Feedback ?
[...]

PR #893, opened 2019-05-15
feat(publish): Support for custom VCL logic

kptdobe:
PR for #812

### DIFF
--- a/src/remotepublish.cmd.js
+++ b/src/remotepublish.cmd.js
@@ -13,6 +13,8 @@
 /* eslint no-unused-vars: ["error", { "argsIgnorePattern": "^_" }] */
 
 const request = require('request-promise-native');
+const fs = require('fs-extra');
+const path = require('path');
 const fastly = require('@adobe/fastly-native-promises');
 const chalk = require('chalk');
 const ProgressBar = require('progress');
@@ -35,6 +37,7 @@ class RemotePublishCommand extends AbstractCommand {
     this._githubToken = '';
     this._updateBotConfig = false;
     this._configPurgeAPI = 'https://app.project-helix.io/config/purge';
+    this._vcl = null;
   }
 
   tick(ticks = 1, message, name) {
@@ -155,6 +158,30 @@ class RemotePublishCommand extends AbstractCommand {
     return this;
   }
 
+  withCustomVCL(value) {
+    if (value.length > 0) {
+      const vcl = {};
+      value.forEach((file) => {
+        try {
+          const fullPath = path.resolve(file);
+          const name = path.basename(fullPath, '.vcl');
+          const content = fs.readFileSync(fullPath).toString();
+          vcl[name] = content;
+        } catch (error) {
+          this.log.error(`Cannot read provided custom vcl file ${file}`);
+          throw error;
+        }
+      });
+      this.withVCL(vcl);
+    }
+    return this;
+  }
+
+  withVCL(value) {
+    this._vcl = value;
+    return this;
+  }
+
   showNextStep(dryrun) {
     this.progressBar().terminate();
     if (dryrun) {
@@ -243,6 +270,7 @@ ${e}`);
         service: this._fastly_namespace,
         token: this._fastly_auth,
         version: this._version,
+        vcl: this._vcl,
       },
     }).then(() => {
       this.tick(9, 'set service config up for Helix', true);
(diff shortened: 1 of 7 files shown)


src/remotepublish.cmd.js
```diff
@@ -155,6 +158,30 @@ class RemotePublishCommand extends AbstractCommand {
     return this;
   }
 
+  withCustomVCL(value) {
+    if (value && value.length > 0) {
```
tripodsan: from an API perspective, I think this should also allow adding single VLCs and update the internal `this._vcl`.
otherwise, rename it to `withCustomVCLs()`

src/remotepublish.cmd.js
```diff
@@ -155,6 +158,30 @@ class RemotePublishCommand extends AbstractCommand {
...
+          this.log.error(`Cannot read provided custom vcl file ${file}`);
+          throw error;
+        }
+      });
+      return this.withVCL(vcl);
+    }
+    return this;
+  }
+
+  withVCL(value) {
```
tripodsan: - no jsdoc
- what's the purpose of this in contrast to `withCustomVCL` ? do you really need both?
- needs to be an object, so name it `withVCLs`

src/remotepublish.cmd.js
```diff
@@ -155,6 +158,30 @@ class RemotePublishCommand extends AbstractCommand {
     return this;
   }
 
+  withCustomVCL(value) {
```
tripodsan: no jsdoc?

test/testRemotePublishCmd.customvcl.js
```diff
@@ -0,0 +1,159 @@
...
+    assert.equal(vcl, null);
+  });
+
+  it('hlx publish --custom-vcl fixtures/vcl/extensions.vcl', async () => {
+    const e = path.resolve(__dirname, 'fixtures/vcl/extensions.vcl');
+    remote.withCustomVCL([e]);
+    await remote.run();
+
+    assert.deepEqual(vcl, {
+      extensions: fs.readFileSync(e).toString(),
```
tripodsan: async, please

test/testRemotePublishCmd.customvcl.js
```diff
@@ -0,0 +1,159 @@
...
+  });
+
+  it('hlx publish --custom-vcl fixtures/vcl/extensions.vcl', () => {
+    const e = path.resolve(__dirname, 'fixtures/vcl/extensions.vcl');
+    const remote = new RemotePublishCommand()
+      .withCustomVCL([e]);
+
+    // eslint-disable-next-line no-underscore-dangle
+    assert.deepEqual(remote._vcl, {
+      extensions: fs.readFileSync(e).toString(),
```
tripodsan: async, please

```suggestion
      extensions: await fs.readFile(e, 'utf8'),
```

test/testRemotePublishCmd.customvcl.js
```diff
@@ -0,0 +1,159 @@
...
+
+  it('hlx publish --custom-vcl fixtures/vcl/extensions.vcl --custom-vcl fixtures/vcl/another.vcl', () => {
+    const e = path.resolve(__dirname, 'fixtures/vcl/extensions.vcl');
+    const a = path.resolve(__dirname, 'fixtures/vcl/another.vcl');
+    const remote = new RemotePublishCommand()
+      .withCustomVCL([e, a]);
+
+    // eslint-disable-next-line no-underscore-dangle
+    assert.deepEqual(remote._vcl, {
+      extensions: fs.readFileSync(e).toString(),
```
tripodsan: async please

src/remotepublish.cmd.js
kptdobe: ah ah. I adopted the jsdoc pattern of that file :) but you are right, that's not a reason, we need to start somewhere.

src/remotepublish.cmd.js
kptdobe: Since the parameter `custom-vcl` (list of file paths) has to be transformed into an object (name:contentoffile), I initially thought we could have the 2 methods to have the 2 ways of providing the data: either the list of files or the object directly. But you are right, this is confusing and introduce useless features. I removed the second method.

kptdobe:
Yeah... managed to bring the code coverage back to positive!

PR #946, opened 2019-06-05
feat(up): checkout and use helix-pages if no source scripts are present

### LINKED ISSUE #929: hlx up without config should render with helix-pages code
when running `hlx up` with no `helix-config.yaml` nor a `src` directory, the project should render using the fallback code from https://github.com/adobe/helix-pages.

The code should be provided via an npm package, so that it is stable in respect to semver when installing `helix-cli`, but also allows to update if needed. alternatively, `hlx up` could automatically download the code from github.

### Suggestion
<del>
- [ ] create a `@adobe/helix-pages` npm package that contains the `src` and `htdocs` folder.
- [ ] add `@adobe/helix-pages` as `dependency` to `helix-cli`
- [ ] `hlx up` will check if a fallback code repository is needed, and if so, it will ensure that the packaged code is used for rendering. eg:
    1. set the `code` and `static` repository to `${hlx-cli-installpath}/node_modules/@adobe/helix-pages`
    2. set the source directory to `${hlx-cli-installpath}/node_modules/@adobe/helix-pages/src`
</del>

- [ ] `hlx up` will check if a fallback code repository is needed, and if so, it will ensure that the
helix-pages repository is checked out to `.hlx/pages/master` and the code is used for rendering. eg:
    1. set the `static` repository to `.hlx/pages/master`
    2. set the source directory to `.hlx/pages/master/src`

tripodsan:
fixes #929


tripodsan:
> hummmm... why copying the whole `helix-pages` project as is ? I think this is weird...

do you mean in the tests? I want to be able to run the test offline:
https://github.com/adobe/helix-cli/blob/0c09cda35ace5f58a4e4a5f4a3ec77b8c124ef87/test/testUpCmdPages.js#L42-L45

> Why dumping `scrani.js` especially if you do nothing with it...

well, it's part of helix-pages at the moment.

> I would simply create a default basic `html.htl`, `html.pre.js` and minimal `style.css` since this is the only thing you want to test for now.

ok. I remove it.

test/fixtures/helix-pages/master/src/html.htl
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>${content.title}</title>
+    <link rel="stylesheet" href="/style.css"/>
+<meta name="viewport" content="width=device-width, initial-scale=1"/>
+<link rel="shortcut icon" type="image/x-icon" href="/favicon.ico">
+<script src="https://d6uhzlpot4xwe.cloudfront.net/runtime/1.21/themes/wiper-fonts.gz.js"></script>
+<script src="/scrani.js"></script>
```
kptdobe:
```suggestion
```

test/fixtures/helix-pages/master/src/html.htl
```diff
@@ -0,0 +1,33 @@
...
+<script>
+  window.onload = function() {
+    scrani.onload();
```
kptdobe:
```suggestion
```

tripodsan:
> I think now it is worst...

ok ok ....

tripodsan:
@kptdobe I initially thought, that we could add more and more tests, to see if the html is rendered correctly. but now I think this should be done in the example project instead. but you're right. I remove the extra files.

PR #983, opened 2019-06-11
Port VCL features to js

koraa:

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/openwhisk/server.js
+++ b/src/openwhisk/server.js
@@ -9,55 +9,15 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
-const fs = require('fs-extra');
-const path = require('path');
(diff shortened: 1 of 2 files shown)


tripodsan:
@koraa I would move all this to a separate action/service (including the changes in the pipeline). having the _server_ logic in helix-cli is IMO not the correct place. similar to the `static.js` that is moved to https://github.com/adobe/helix-static, we could have a `helix-resolver` or `helix-dispatcher` (I try to avoid the term _server_ here :-).

trieloff:
I fully agree with @tripodsan and this is similar to what I suggested here https://github.com/adobe/helix-pipeline/issues/365#issuecomment-500013338

tripodsan:
I also added some more comments to https://github.com/adobe/helix-pipeline/issues/365#issuecomment-501501045

PR #1008, opened 2019-06-20
feat(deploy): Add RESOLVE_GITREF_SERVICE as action param

### LINKED ISSUE #1007: Set API_RESOLVE_GIT_REF action params during deploy.
the pipeline will support automatic ref lookup using the https://github.com/adobe/helix-resolve-git-ref service. but that that the `API_RESOLVE_GIT_REF` needs to be set to the correct API url of the service during deploy.

tripodsan:
fixes #1007

needs https://github.com/adobe/helix-pipeline/pull/384 to have an effect.

### DIFF
--- a/src/deploy.cmd.js
+++ b/src/deploy.cmd.js
@@ -53,6 +53,7 @@ class DeployCommand extends StaticCommand {
     this._createPackages = 'auto';
     this._addStrain = null;
     this._enableMinify = false;
+    this._resolveGitRefSvc = 'helix-services/resolve-git-ref@v1';
   }
 
   get requireConfigFile() {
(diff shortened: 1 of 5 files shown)


trieloff:
What's with the test coverage drop?

tripodsan:
> What's with the test coverage drop?

hmm....

BTW: I set it to WIP since I wanted to revisit the action param name and also if we should set the http API url or just the action name. The action name has the advantage, that it can be namespace independent, since it is invoked from within the users namespace anyways.

tripodsan:
> What's with the test coverage drop?

there is a new method in `deploy.cmd.js`:
```
  withResolveGitRefService(value) {
    this._resolveGitRefSvc = value;
    return this;
  }
```
which is never tested, since the CLI tests mock all those ones, and we don't set it in integration tests...

trieloff:
> and we don't set it in integration tests

so it's easy to fix.

tripodsan:
>> and we don't set it in integration tests ...

@trieloff not really. there are many other not covered lines that would be difficult to test. we can add a test just to set it, but without measuring the correct outcome it doesn’t make sense. we should not try to cover lines just for the sake of covering. 

anyways, I’ll add it to the deploy integration test later.

PR #1020, opened 2019-06-24
feat(publish): use versioned publish API

trieloff:
The old publish API is still available at https://adobeioruntime.net/api/v1/web/helix/default/publish but won't get any new features

See https://github.com/adobe/helix-publish/pull/117


trieloff:
Approved together with adobe/helix-publish#117

PR #1037, opened 2019-06-28
test(html pipe): Update integration test for newly exposed types in the meta JSON attributes

ramboz:
Adding the newly exposed `types` array on the `meta` object.
 
fix adobe/helix-pipeline#279


ramboz:
@tripodsan Sure, makes sense.

PR #1064, opened 2019-07-05
feat(deploy): must run hlx build

### LINKED ISSUE #241: [hlx deploy] must run hlx build
While setting up my production environment and fixing some bugs, I had to run deploy multiple times. Even though I test with `hlx up` before running the `hlx deploy`, the build is not up-to-date and I keep forgetting to run `hlx build`. Testing on production is pretty slow, thus I keep losing time understanding why my changes did not have the expected result. The only reason is that the code is not live... Error prone and quite frequent.

`hlx deploy` should either run a `hlx build` for me. Telling me the build is not up to date would not be a good option because I never want to deploy an older version... (I could still checkout the old code if really needed)

N/A

rofe:
Fixes #241 (see issue for implementation details)


src/yargs-build.js
```diff
@@ -24,7 +24,7 @@ module.exports = function commonArgs(yargs) {
       type: 'string',
     })
     // allow for comma separated values
-    .coerce('files', value => value.reduce((acc, curr) => {
+    .coerce('files', value => (typeof value !== 'object' ? [value] : value).reduce((acc, curr) => {
```
tripodsan: maybe better check with `Array.isArray(value)` ?

rofe: Right

PR #1103, opened 2019-07-25
Continuous Integration on Windows

trieloff:
There are more than 16 test failures right now, many seem path-related or otherwise Windows-specific, so I'll need a windows VM to fix them.

### DIFF
--- a/test/testGitUtils.js
+++ b/test/testGitUtils.js
@@ -25,7 +25,9 @@ git.plugins.set('fs', require('fs'));
 
 const GitUtils = require('../src/git-utils');
 
-const GIT_USER_HOME = path.resolve(__dirname, 'fixtures/gitutils');
+const GIT_USER_HOME = path.resolve(__dirname, 'fixtures', 'gitutils');
+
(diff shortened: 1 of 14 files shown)


test/testGitUtils.js
```diff
@@ -82,7 +82,12 @@ describe('Testing GitUtils', () => {
     assert.equal(await GitUtils.isDirty(testRoot, GIT_USER_HOME), true);
   });
 
-  it('isDirty #unit with submodules', async () => {
+  it('isDirty #unit with submodules', async function test() {
```
trieloff: https://github.com/adobe/helix-testutils can help with that.

tripodsan: my IDE doesn't llike the `condit`... ;-)

tripodsan: but the problem is, when the test is skipping itself, the `afterEach` is not executed anymore; which can lead to side effects...

tripodsan: see https://github.com/mochajs/mocha/issues/3740

Issue #1111, opened 2019-08-06
Demo, Fails if .gitconfig doesn't exist, but says Git isn't installed.


MarquiseRosier:
**Description**
When running the demo; hlx demo [demo_name] fails if .gitconfig doesn't exist.

**To Reproduce**
1) Move .gitconfig out of home directory. To figure out what dir this is, fire node and see what os.homedir() prints. 

2) cd to helix-cli root, and run hlx demo [demo_name], it will fail and say to install Git

**Expected behavior**
Helix CLI should know that Git is installed, but a .gitconfig isn't there! 

**Screenshots**
![Screen Shot 2019-08-06 at 2 34 11 PM](https://user-images.githubusercontent.com/12217245/62579438-9badc400-b858-11e9-8e3e-0d22c11db26e.png)

**Version:**
4.9.10

**Additional context**
Super easy fix, just convenience things!

stefan-guggisberg:
nitpicking:

The title of this issue states 

> but says Git isn't installed.

The error message however says

> It seems like Git has not yet been setup on this system. 

and the provided link refers to [First Time Git Setup](https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup).

There's a difference between [Installing Git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git) and [First Time Git Setup](https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup) as illustrated by the 2 separate chapters in the Git documentation.

The error message and documentation link are concise and correct.

PR #1118, opened 2019-08-12
Issues/cli/1111

MarquiseRosier:
Issue: 
https://github.com/adobe/helix-cli/issues/1111

Added check for error code 127 -> command doesn't exist. 

Propagated Errors for this particular ^ Error for user clarity.

### DIFF
--- a/src/demo.cmd.js
+++ b/src/demo.cmd.js
@@ -46,8 +46,6 @@ class InitCommand {
       shell.exec(cmd, (code, stdout, stderr) => {
         if (code === 0) {
           resolve(code);
-        } else if (code === 127) {
-          resolve(code);
         } else {
           reject(stderr);
         }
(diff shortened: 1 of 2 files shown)


src/demo.cmd.js
```diff
@@ -100,12 +103,24 @@ class InitCommand {
...
     // #181 cover edge case: make sure git is properly configured
-    if (!await fse.pathExists(`${os.homedir()}/.gitconfig`)) {
-      throw new Error(`
-It seems like Git has not yet been setup on this system. 
-
-See https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup for more information.  
-`);
+    try {
+      if (await this.execAsync('git --version') === 0) {
+        if (!await fse.pathExists(`${os.homedir()}/.gitconfig`)) {
```
trieloff: You should use `path.resolve` instead of string concatenation. This might also help with the Windows issues.

test/testDemoCmd.js
```diff
@@ -35,6 +36,33 @@ describe('Integration test for demo command', function suite() {
...
+    assert.doesNotReject(demoInstance
+      .withDirectory(testDir)
+      .withName('project1')
+      .withType('full')
+      .run());
+  });
+
+  it('test for failure, when Git is not installed', async () => {
+    const demoInstance = new DemoCommand();
+    sinon.stub(demoInstance, 'execAsync').returns(127);
```
trieloff: Instead of stubbing the `execAsync` command, it might be better to have a Mocha `before` and `after` that moves `.gitconfig` aside and then back after the test. It should be relatively easy to recreate the test condition and it will help us detect issues in case the `git` return status code changes, etc.

src/demo.cmd.js
```diff
@@ -52,6 +40,21 @@ class InitCommand {
     this._type = 'simple';
   }
 
+  // eslint-disable-next-line class-methods-use-this
```
trieloff: I think there is no good reason to move the `execAsync` function into the class. The fact that you had to disable an ESLint check is a good warning sign. I think it has been caused largely by the need to easily stub the method, but given that creating real test conditions is easy, I'd go down that route. (see below)

MarquiseRosier: Good point! Hmmm, you're right. 

I'm torn, we can either keep the stub approach; or manipulate the environment by doing things like removing the .gitconfig in the before() and restoring it in the after(). 

If we move execAsync outside of the InitCommand class; I believe there's no way to stub it; such that the invocations inside of InitCommand are rerouted. 

@tripodsan and @trieloff what will be the choice? @tripdsan approach would keep us from doing things like moving;installing git as we can just stub the execAsync and make it return failure/success...for people who want to run the tests locally. 

Or we can use @trieloff approach; where we trade stubbing for actually creating the conditions, even when tests are local. 

Whatdya think fellas?

Issue #1112, opened 2019-08-07
hlx publish command rename


MarquiseRosier:
**Description**
hlx publish command needs service id not namespace: --fastly-namespace <service-ID>

**To Reproduce**
Make a tiny change in a MD, commit and publish to github
deploy the project, hlx deploy --fastly-namespace <namespace>
publish project , hlx publish --fastly-namespace <service-ID> 

**Expected behavior**
parameter name should reflect what command needs.

**Screenshots**
![Screen Shot 2019-08-07 at 3 00 17 PM](https://user-images.githubusercontent.com/12217245/62660834-17287780-b924-11e9-9ccd-6895e3450722.png)

**Version:**
4.9.10

**Additional context**
Nada!

tripodsan:
so we should probably name it `--fastly-serviceid`

trieloff:
Please make sure that the old name stays as an alias, so that we don't break compatibility.

MarquiseRosier:
Got it!

PR #1120, opened 2019-08-14
refactor(cli): rename Option

MarquiseRosier:
Rename fastly-namespace to fastly-serviceid; as that is what is being requested

"fix 1112"

https://github.com/adobe/helix-cli/issues/1112## Related Issues

Thanks for contributing!

### DIFF
--- a/src/publish.js
+++ b/src/publish.js
@@ -13,7 +13,6 @@
 'use strict';
 
 const yargsOpenwhisk = require('./yargs-openwhisk.js');
-const yargsFastly = require('./yargs-fastly.js');
 const yargsGithub = require('./yargs-github.js');
 const { makeLogger } = require('./log-common.js');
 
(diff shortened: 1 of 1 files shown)


src/publish.js
```diff
@@ -13,7 +13,6 @@
 'use strict';
 
 const yargsOpenwhisk = require('./yargs-openwhisk.js');
-const yargsFastly = require('./yargs-fastly.js');
```
tripodsan: I suggest to rename the option in `yargs-fastly.js`, so it is consistent for all commands that need fastly credentials.

MarquiseRosier: I slacked you about this! :)

MarquiseRosier: this command doesn't need fastly-namespace
if we use yargs-fastly it'll attach fastly-namespace as an option

but Lars wants fastly-namespace to be an alias for fastly-serviceid

There will be a collision....fastly-namespace will be both an option and an alias

MarquiseRosier: don't even know if that's possible. But it can lead to things like; someone can pass both --fastly-serviceid and --fastly-namespace and if both are allowed to exist... --fastly-namespace can overwrite

tripodsan: no, `fastly-namespace` was an unfortunate naming of the parameter for the `serviceid`.
so we should use the `fastly-serviceid` parameter instead.

this parameter is used in `publish.js` and in `deploy.js`. so we should keep it consistent.
you can add an alias using https://github.com/yargs/yargs/blob/master/docs/api.md#aliaskey-alias

eg: 

`alias: 'fastly-namespace',`

PR #1129, opened 2019-08-21
feat(cli): allows Override of Gateway Timeout

MarquiseRosier:
You can now supply a host of action parameters via key-value pairs. Supplying --action-param
HTTP_TIMEOUT 2000 will ensure to replace the default timeout of 1000.

"feat 911"

Relies on: https://github.com/adobe/helix-simulator/tree/issues/cli/911

## Related Issues

Thanks for contributing!

### DIFF
--- a/demos/full/_env
+++ b/demos/full/_env
@@ -8,5 +8,5 @@
 #HLX_FASTLY_AUTH=
 #HLX_FASTLY_NAMESPACE=
 
-# Github Bot
+# Github Bot & authenticated GitHub access
 #HLX_GITHUB_TOKEN=
--- a/src/remotepublish.cmd.js
+++ b/src/remotepublish.cmd.js
@@ -315,6 +315,10 @@ ${e}`);
       const namespace = this._fastly.writeDictItem(this._version, 'secrets', 'OPENWHISK_NAMESPACE', this._wsk_namespace);
       jobs.push(namespace);
     }
+    if (this._githubToken) {
+      const token = this._fastly.writeDictItem(this._version, 'secrets', 'GITHUB_TOKEN', this._githubToken);
+      jobs.push(token);
+    }
     return Promise.all(jobs).then(() => {
       this.tick(2, 'enabled authentication', true);
       return true;
--- a/test/testPublishCli.js
+++ b/test/testPublishCli.js
@@ -97,7 +97,7 @@ describe('hlx publish', () => {
     sinon.assert.calledOnce(mockPublish.run);
   });
 
-  it('hlx publish works with minimal arguments with serviceid', () => {
+  it('hlx publish handles serviceid', () => {
     new CLI()
       .withCommandExecutor('publish', mockPublish)
       .run(['publish',
@@ -107,14 +107,7 @@ describe('hlx publish', () => {
         '--fastly-serviceid', 'hlx',
       ]);
 
-    sinon.assert.calledWith(mockPublish.withWskHost, 'adobeioruntime.net');
-    sinon.assert.calledWith(mockPublish.withWskAuth, 'secret-key');
-    sinon.assert.calledWith(mockPublish.withWskNamespace, 'hlx');
-    sinon.assert.calledWith(mockPublish.withFastlyNamespace, 'hlx'); // TODO !!
-    sinon.assert.calledWith(mockPublish.withFastlyAuth, 'secret-key');
-    sinon.assert.calledWith(mockPublish.withCustomVCLs, []);
-    sinon.assert.calledWith(mockPublish.withDispatchVersion, undefined);
-    sinon.assert.calledOnce(mockPublish.run);
+    sinon.assert.calledWith(mockPublish.withFastlyNamespace, 'hlx');
   });
 
   it('hlx publish implicit bot config with github token', () => {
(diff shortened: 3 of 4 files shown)


src/up.js
```diff
@@ -66,6 +66,31 @@ module.exports = function up() {
           type: 'int',
           default: 3000,
         })
+        .option('action-param', {
```
trieloff: The code below is doing the same (conceptually) as the code here: https://github.com/adobe/helix-cli/blob/07ba48b0b9f8fdaf92c8eed7e7ada047f15676e3/src/deploy.js#L99-L107

In `hlx deploy --default key1 value1 --default key2 value2` is the same as `hlx up --action-param key1 value1 key2 value2`. I'd suggest we unify these two parameters by
- [ ] moving the code from `hlx deploy` into a `yargs-params.js` file
- [ ] adding `action-param` as an alias to defaults
- [ ] using the `yargs-params.js` function in both `hlx up` and `hlx deploy`

trieloff:
Also, I'm not sure what the `break()` message should imply, as it's not part of the conventional commits spec: https://www.conventionalcommits.org/en/v1.0.0-beta.4/#specification and there is no breaking change (you are making additive changes only)

I'd assume a commit message like:

```
feat(up): enable specification of default parameters for actions

fixes #911 among others
```

src/up.js
tripodsan: I'm not sure this should be treated the same. you have different requirements, eg. for `HTTP_TIMEOUT` on your local dev instance than on production. since you can add the params to you `.env` files, this might add unexpected params during deployment.

we could rename it to `--dev-default` to align it closer to the deploy arguments.

MarquiseRosier: good idea @trieloff i'll do that. Do you think I should keep the approach from hlx up for checking params; I think it's more flexible to do 
--default KV KV  instead of --default KV --default KV, also the code in deploy assumes string values, which will break if you supply HTTP_TIMEOUT 2000; i'll just check for string

@tripodsan i'll keep your naming convention; I will keep --defualt as the primary; and then alias --dev-default.

tripodsan: @MarquiseRosier I would not use `--default` for `hlx up`. As I said, it causes confusion when setting the parameters in `.env`.

MarquiseRosier: This is true. Hmmm I definitely want to make a yargs-params.js for this code; So i'm going to put two parameters in commonArgs(yargs, name, describe, alias, type, default) so we can reuse logic here

src/up.js
```diff
@@ -27,6 +28,13 @@ module.exports = function up() {
     description: 'Run a Helix development server',
     builder: (yargs) => {
       yargsBuild(yargs);
+      yargsParams(yargs, {
+        name: 'dev-default',
+        describe: 'Additional action parameters',
+        alias: ['action-params', 'actionParams', 'devDefault'],
```
tripodsan: I think we don't need the `acton-param` and `actionParams` aliases anymore.

test/testUpCli.js
```diff
@@ -54,6 +55,12 @@ describe('hlx up', () => {
     sinon.assert.calledOnce(mockUp.run);
   });
 
+  it('hlx up can override secrets', () => {
+    new CLI()
+      .withCommandExecutor('up', mockUp)
+      .run('up --dev-default HTTP_TIMEOUT 2000');
+  });
```
tripodsan: where is the assert that the param was set?

MarquiseRosier: Good point! Yeah I definitely duplicated logic. Thank you for pointing that out!

src/up.js
MarquiseRosier: Ok, I'll erase it :D

src/up.js
trieloff: > I think it's more flexible to do `--default KV KV` instead of `--default KV --default KV`

That might be, but I'd take the default that is built in with yargs.

src/yargs-params.js
```diff
@@ -0,0 +1,35 @@
...
+      // Get rid of name as we already consumed it
+      const optCopy = { ...options };
+      delete optCopy.name;
+      return optCopy;
+    })())
+    .coerce((options.name), (argv) => {
+      // Check Length is Multiple of 2 these are KV pairs.
+      const { length } = argv;
+      const retMap = {};
+      if (length > 0 && length % 2 === 0) {
```
trieloff: There should be a warning/error if the length is not a multiple of 2.

src/yargs-params.js
```diff
@@ -0,0 +1,35 @@
...
+        const numberOfArgs = length / 2;
+        for (let idx = 0; idx < numberOfArgs; idx += 1) {
+          const key = argv[2 * idx].toUpperCase();
```
trieloff: Not sure about the enforced upper-casing. It's very non-obvious.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@adobe/helix-cli",
-  "version": "5.0.0",
+  "version": "5.0.1",
```
tripodsan: why is the package.json modified?

MarquiseRosier: CLI repo changed it already; so it became merge conflict!

tripodsan: then I don't understand the diff here. can you rebase everything on master?

MarquiseRosier: good point! I should be able to.

MarquiseRosier: You know of a way to rebase and keep history?

MarquiseRosier: wait nevermind, thanks Tobi :)

Issue #901, opened 2019-05-20
Support authenticated GitHub access


stefan-guggisberg:
**Is your feature request related to a problem? Please describe.**
Currently content is fetched from VCL (Fastly) & serverless actions (AdobeIO Runtime) via unauthenticated GitHub raw requests. As a result private GitHub repos are not accessible. Furthermore unauthenticated GitHub raw requests might become subject to potential rate limiting/throttling.

**Describe the solution you'd like**
If there's a `HLX_GITHUB_TOKEN` environment variable `hlx publish` could add it to the `secrets` edge dictionary, thus make it available to custom VCL and AdobeIO Runtime calls (via parameter).

**Related issues** 
 - adobe/helix-cli#901 (`hlx publish --github-token`)
 - adobe/helix-publish#165
 - adobe/helix-pipeline#72 
 - adobe/helix-dispatch#71
 - adobe/helix-resolve-git-ref#1
 - adobe/helix-static#46

rofe:
How would we handle the use case where both code and content repos are private, but require different access tokens (for example a company would like to have all their subsites rendered using the same code, which is under the web team’s control, while different marketing teams maintain their content repos and publish them using the common code repo)?

PS: The issue title describes the implementation, not the feature, which is: support private GitHub repositories ;)

stefan-guggisberg:
> How would we handle the use case where both code and content repos are private, but require different access tokens (for example a company would like to have all their subsites rendered using the same code, which is under the web team’s control, while different marketing teams maintain their content repos and publish them using the common code repo)?

Do you think this is a common use case? If not I would prefer to first address the simple use case first.

> PS: The issue title describes the implementation, not the feature, which is: support private GitHub repositories ;)

Feel free to change it ;) The initial feature I had in mind was supporting authenticated GitHub access to alleviate potential rate limiting/throttling issues.

tripodsan:
we already use a `--github-token` (https://github.com/adobe/helix-cli/blob/master/src/publish.js#L94) which is obtained via `hlx auth` and used to access by the bot to access to repository to update the purge config.

I'm not sure if it will not cause confusions:

(assuming the token is added via `HLX_GITHUB_TOKEN` or passed via `--github-token`)

case 1: bot **not** installed, the user created a personal access token
- `hlx publish` will tell, that they should install the bot ✅
- _fetch_ can read the private repo ✅

case 2: bot installed, the user created a personal access token
[...]

PR #1138, opened 2019-08-28
publish: write Github token to Fastly service dictionary

stefan-guggisberg:
## Related Issues
#901

### DIFF
--- a/demos/full/_env
+++ b/demos/full/_env
@@ -8,5 +8,5 @@
 #HLX_FASTLY_AUTH=
 #HLX_FASTLY_NAMESPACE=
 
-# Github Bot & private repo support
+# Github Bot & authenticated GitHub access
 #HLX_GITHUB_TOKEN=
--- a/demos/full/_env
+++ b/demos/full/_env
@@ -8,5 +8,5 @@
 #HLX_FASTLY_AUTH=
 #HLX_FASTLY_NAMESPACE=
 
-# Github Bot
+# Github Bot & private repo support
 #HLX_GITHUB_TOKEN=
--- a/src/publish.js
+++ b/src/publish.js
@@ -84,10 +84,7 @@ module.exports = function strain() {
           'Fastly Service ID is required.',
         )
         .check((args) => {
-          if (args.githubToken && args.updateBotConfig === undefined) {
-            // eslint-disable-next-line no-param-reassign
-            args.updateBotConfig = true;
-          } else if (args.updateBotConfig && !args.githubToken) {
+          if (args.updateBotConfig && !args.githubToken) {
             return new Error('Github token is required in order to update bot config.\n'
               + 'Provide one via --github-token or via the HLX_GITHUB_TOKEN environment variable.\n'
               + 'You can use `hlx auth` to automatically obtain a new token.');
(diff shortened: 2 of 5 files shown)


demos/full/_env
```diff
@@ -8,5 +8,5 @@
 #HLX_FASTLY_AUTH=
 #HLX_FASTLY_NAMESPACE=
 
-# Github Bot
+# Github Bot & private repo support
```
tripodsan: hmmm...I'm not sure if we can use the same token for both...we need to try it out...

src/publish.js
```diff
@@ -84,10 +84,7 @@ module.exports = function strain() {
           'Fastly Service ID is required.',
         )
         .check((args) => {
-          if (args.githubToken && args.updateBotConfig === undefined) {
-            // eslint-disable-next-line no-param-reassign
-            args.updateBotConfig = true;
-          } else if (args.updateBotConfig && !args.githubToken) {
+          if (args.updateBotConfig && !args.githubToken) {
```
tripodsan: hmm...you changed the behaviour for automatically updating the bot config... why?

test/testPublishCli.js
```diff
@@ -107,34 +107,10 @@ describe('hlx publish', () => {
...
-    sinon.assert.calledWith(mockPublish.withWskNamespace, 'hlx');
-    sinon.assert.calledWith(mockPublish.withFastlyNamespace, 'hlx'); // TODO !!
-    sinon.assert.calledWith(mockPublish.withFastlyAuth, 'secret-key');
-    sinon.assert.calledWith(mockPublish.withCustomVCLs, []);
-    sinon.assert.calledWith(mockPublish.withDispatchVersion, undefined);
-    sinon.assert.calledOnce(mockPublish.run);
+    sinon.assert.calledWith(mockPublish.withFastlyNamespace, 'hlx');
   });
 
-  it('hlx publish implicit bot config with github token', () => {
```
tripodsan: why s this removed?

src/publish.js
stefan-guggisberg: Because `githubToken` is not used exclusively for configuring the bot but also for supporting authenticated github access. The implicit bot configuration when providing a github token was just a convenience feature and is confusing now that the github token is also used for authenticated github access.

test/testPublishCli.js
stefan-guggisberg: see my previous comment above.

demos/full/_env
stefan-guggisberg: I agree. OTOH having to specifiy separate github tokens would be counter-intuitive and confusing.

src/publish.js
tripodsan: I disagree. bot config should always be updated when a github token is present. the previous behaviour was correct and should be retained.

test/testPublishCli.js
```diff
@@ -201,17 +208,16 @@ describe('hlx publish', () => {
     sinon.assert.calledWith(mockPublish.withDispatchVersion, 'ci1');
   });
 
-  it('hlx publish handles dispatch-version with serviceid', () => {
+  it('hlx publish handles fastly-serviceid', () => {
```
tripodsan: why is this changed?

demos/full/_env
tripodsan: ok..let's see how this all works out

PR #1147, opened 2019-09-09
Introduce Option for Soft and Hard Purges

trieloff:
See https://github.com/adobe/helix-home/issues/51

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,14 @@ Publishing [========================================----------]  4.1s
 ✅  All strains have been published and version 89 is now online.
 ```
 
+### Purging the Cache upon Publishing
+
+Whenever you run `hlx publish`, a new version of your site, with potentially changed code will be made available to visitors. For visitors to see the changes, the Fastly cache needs to be purged. By default, `hlx publish` uses a "Soft purge", which means that the entire content of your website will be marked as outdated (or stale), but not actually removed from the cache. When a request for a cached file that has been marked outdated hits Fastly, Fastly will serve the old version, but fetch a new version in the background. As a result, your site is still as fast as before, *but in order for changes to show up, two requests are needed*.
+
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -112,6 +112,14 @@ Publishing [========================================----------]  4.1s
...
+
+If you want to see your changes faster, at the expense of slower load times right after publishing, use the command `hlx publish --purge hard`, which triggers a hard purge, i.e. removes all cached objects from the Fastly CDN. Doing this on a site with substantial traffic is unwise, but it can be a useful option during development.
+
+Finally, if you do not want the cache to be purged at all, run `hlx publish --skip`. Your changes will only become visible when the cached objects expire or the cache is cleared in some other way, for instance from the Fastly console or using an API call as part of a more compex continuous deployment set-up.
```
tripodsan:
```suggestion
Finally, if you do not want the cache to be purged at all, run `hlx publish --purge skip`. Your changes will only become visible when the cached objects expire or the cache is cleared in some other way, for instance from the Fastly console or using an API call as part of a more compex continuous deployment set-up.
```

src/remotepublish.cmd.js
```diff
@@ -237,20 +243,26 @@ class RemotePublishCommand extends AbstractCommand {
   }
 
   purgeFastly() {
-    if (this._dryRun) {
+    if (this._dryRun || !(this._purge === 'soft' || this._purge === 'hard')) {
```
tripodsan:
```suggestion
    if (this._dryRun || this._purge === 'skip') {
```

trieloff: Your code is easier, but it use the hard purge as the default fallback in case the publish command is ever called with an invalid state. I prefer the non-destructive default.

tripodsan: but the condition is very simple: either it's a dry-run or purging is skipped, then skip.
whether or not the purge is `soft` or  `hard`, doesn't matter in this case.

if you later add a new option: `--purge fluffy`, you need to update this condition.

Issue #1143, opened 2019-09-05
[publish] error if github token is present


rofe:
**Description**
`hlx publish` throws an error during publishing with `--github-token=...`

**To Reproduce**
1. Run `hlx publish --github-token=...` on your demo project
2. See error
```
Publishing [==============================================----] updating helix-bot purge config 2.1serror: Error while running the Publish command: StatusCodeError: 403 - undefined
    at new StatusCodeError (/Users/foo/code/helix-cli/node_modules/request-promise-core/lib/errors.js:32:15)
    at Request.plumbing.callback (/Users/foo/code/helix-cli/node_modules/request-promise-core/lib/plumbing.js:104:33)
    at Request.RP$callback [as _callback] (/Users/foo/code/helix-cli/node_modules/request-promise-core/lib/plumbing.js:46:31)
    at Request.self.callback (/Users/foo/code/helix-cli/node_modules/request/request.js:185:22)
    at Request.emit (events.js:198:13)
    at Request.<anonymous> (/Users/foo/code/helix-cli/node_modules/request/request.js:1161:10)
    at Request.emit (events.js:198:13)
    at IncomingMessage.<anonymous> (/Users/foo/code/helix-cli/node_modules/request/request.js:1083:12)
    at Object.onceWrapper (events.js:286:20)
    at IncomingMessage.emit (events.js:203:15)403 - undefined
```

**Expected behavior**
No error is throw during publishing

**Version:**
5.0.2

**Additional context**
I ran into this while testing the new private repository support.

MarquiseRosier:
Hmmm; you want to suppress the error; or ignore it and return to publishing without updateBotConfig?

rofe:
I guess we could... however, we should also understand why the response is forbidden...

MarquiseRosier:
Good point;  I know there's a request sent to https://apps.project-helix.io/config/purge and when the github_token is something obviously wrong; it throws the 403; i'm guessing it has to do with not having access to the repo because of bad token

rofe:
The token is valid, private repo access is working. But maybe it's because I created it manually and not using `hlx auth`? 🤷‍♂

MarquiseRosier:
Hmmm ^^ agreed; another question is where are you generating the token from?

rofe:
https://github.com/settings/tokens

tripodsan:
As I said earlier, those are probably 2 different tokens. I'll look into it.

MarquiseRosier:
[...]

PR #1148, opened 2019-09-10
Bugs/cli/1143

MarquiseRosier:
https://github.com/adobe/helix-cli/issues/1143

### DIFF
--- a/src/remotepublish.cmd.js
+++ b/src/remotepublish.cmd.js
@@ -481,7 +481,7 @@ The provided GITHUB_TOKEN is not authorized to act on behalf
 of the Helix Bot and can therefore not be used to update the purge config. 
 You can generate a new token by running 'hlx auth'`);
       } else {
-        this.log.error(`${message}: ${e.stack}`);
+        this.log.error(`${message}: ${e.stack}`, e);
         throw new Error(message, e);
       }
     }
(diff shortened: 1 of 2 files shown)


src/remotepublish.cmd.js
```diff
@@ -475,8 +476,15 @@ ${e}`);
       this.showNextStep(this._dryRun);
     } catch (e) {
       const message = 'Error while running the Publish command';
-      this.log.error(`${message}: ${e.stack}`, e);
-      throw new Error(message, e);
+      if (e.statusCode === 403) {
+        throw new Error(`${message}:
+The provided GITHUB_TOKEN is not authorized to act on behalf
```
rofe: @MarquiseRosier: is the token really the actor here? I think this should probably be rephrased to sth like _The provided GITHUB_TOKEN is not authorized to be used by the Helix Bot to update the purge config._

tripodsan: the wording is technically correct:
see https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps/#authenticating-as-an-installation

rofe: Maybe I'm missing something, but I still think the actor in the sentence should be the bot, not the token which allows the bot to act... anyway, I forgot to add the _nitpicking_ label ;)

MarquiseRosier: Ahh snap that's a good point haha. I'll definitely submit a PR for another issue; and lump the rename in there! I'll get a review from you :D :)

PR #1151, opened 2019-09-17
feat(pub): add support for debug key

MarquiseRosier:
https://github.com/adobe/helix-publish/issues/38


src/publish.js
```diff
@@ -14,6 +14,7 @@
 
 const yargsOpenwhisk = require('./yargs-openwhisk.js');
 const yargsFastly = require('./yargs-fastly.js');
+const yargsDebug = require('./yargs-debug.js');
```
tripodsan: I don't think it is necessary to create an extra file for this option, as it can only be used during `hlx publish`.

MarquiseRosier: Yeah this is a good point! @kptdobe raised this point earlier; I will remove this

trieloff:
@MarquiseRosier you should also update this documentation: https://github.com/adobe/project-helix.io/blob/master/doc/general/troubleshooting.md – and maybe move it to `helix-home`

PR #1183, opened 2019-10-15
feat(publish): warn if remote logging fails

### LINKED ISSUE #890: hlx publish should warn, not fail when helix-logging fails
As the [ongoing incident](https://status.project-helix.io/incidents/jl69shj426hs) means that `helix publish` will be affected for new sites, I suggest we warn instead of fail when the `helix-logging` service cannot be used.

medusalix:

## Related Issues

Resolves #890

Thanks for contributing!


src/remotepublish.cmd.js
```diff
@@ -244,7 +244,7 @@ class RemotePublishCommand extends AbstractCommand {
     }).catch((e) => {
       this.tick(10, 'failed to set up logging', true);
       this.log.error(`Remote addlogger service failed ${e}`);
-      throw new Error('Unable to set up remote logging');
+      this.log.warn('Unable to set up remote logging');
```
tripodsan: I think the `log.error` above is enough. no need to log again with warn. Maybe change the log level to `warn`

PR #1198, opened 2019-10-30
feat(build): create transparent builds

### LINKED ISSUE #1197: Create transparent builds
### Is your feature request related to a problem? Please describe.
The way helix cli build the dynamic actions result in some javascript files that makes it hard to understand by the developer how the pipeline is executed. This also poses a problem when trying to debug the pipeline.

### Describe the solution you'd like
Ideally, helix cli should produce a more or less self-contained project that has well defined entry points and module dependencies.

### Current layout

in a simple case, the `src` and `.hlx/build` directory look llike:

```
src/
├── html.htl
└── html.pre.js
.hlx/
└── build/
    ├── html.info.json
    ├── html.js
    ├── html.js.map
    ├── html.pre.info.json
    ├── html.pre.js
    └── html.pre.js.map
```

where the `html.js` is a concatenation of:
- the compiled HTL script
- a wrapper that sets up the pipeline
- the openwhisk action entry point

The imports for `htlengine` and `pipeline` are somehow magically resolved using the module scope of `helix-cli`.

### Suggestion

- the build directory should be self contained if possible. maybe having a `package.json` and `node_modules`
- instead of concatenation, use imports.
- install `helix-pipeline` and `htlengine` on demand or copy them from the `helix-cli/node_modules`. the bare installation of those currently uses **56mb** and has **274** dependencies.

```
.hlx/build/
├── html.htl.js
├── html.js
├── html.pre.js
├── html.pipe.js
├── node_modules/
├── package-lock.json
└── package.json
```

**Discuss**:
should the `html.pre.js` be copied into the build directory, or should it be required with `../../src/html.pre.js` ?

tripodsan:
fixes #1197 

- removed parcel dependency
- faster builds
- require source files intuitively

---

a simple `html.htl` with a `html.pre.js` is now transformed into:

```
.hlx/
└── build/
    └── src/
        ├── html.info.json
        ├── html.js
        ├── html.script.js
        └── html.script.js.map
```

with a generated:

**html.js**
```
...
const { pipe } = require('@adobe/helix-pipeline/src/defaults/html.pipe.js');
const { pre, before, after, replace } = require('../../../src/html.pre.js');
const script = require('./html.script.js');
...
```

the `html.htl` was transformed into `html.script.js`.

---

in case of a _pure_ script, like `xml.js` with a custom `xml.pipe.js`:

```
.hlx/
└── build/
    └── src/
        ├── xml.info.json
        └── xml.js
```

with a generated:

**xml.js**
```
...
const { pipe } = require('../../../src/xml.pipe.js');
const { pre, before, after, replace } = require('@adobe/helix-pipeline/src/defaults/xml.pre.js');
const script = require('./../../../src/xml.js');
...
```


alexkli:
LGTM. The `../..` is currently a bit tricky for `wskdebug` because it mounts the `.hlx/build` folder into the docker container, where `../` doesn’t work, but maybe an option can be added to handle this (base dir vs. entry point file).

tripodsan:
> LGTM. The `../..` is currently a bit tricky for `wskdebug` because it mounts the `.hlx/build` folder into the docker container, where `../` doesn’t work, but maybe an option can be added to handle this (base dir vs. entry point file).

yes please, i'd rather not copy the source files to the build folder again....

kptdobe:
I assume all the existing tests validate that the build works as before.

tripodsan:
> I assume all the existing tests validate that the build works as before.

yes...some of the tests were modified and the JSX test is skipped. Also, since we don't use parcel anymore, building no longer validates the sources.

tripodsan:
> What would be needed to keep JSX support?

Some sort of JSX processor... but I would (re)add it, once someone asks about it :-)

trieloff:
The JSX processing is done by `babel`, which is still in the dependency tree: https://github.com/adobe/helix-cli/blob/b70d9c7dc32827927fbd870b03e69b12e848235b/package-lock.json#L340

tripodsan:
> The JSX processing is done by `babel`, which is still in the dependency tree:

I cleaned up the dependencies :-) but I'll see what can be done :-)

tripodsan:
@trieloff I added JSX support again.... no more breaking change :-)

Issue #1189, opened 2019-10-22
`hlx demo` & `hlx up` fails


stefan-guggisberg:
**Description**
Starting the local development server with `hlx up` in a new project created with `hlx demo` fails with the following error:
```
Unable to determine origin url for --local-repo = .
```

**To Reproduce**
1. Open terminal and enter
```
$ hlx demo demo
$ cd demo
$ hlx up
Unable to determine origin url for --local-repo = .
```

**Expected behavior**
`hlx demo` follow by `hlx up` should work out of the box

**Version:**
```bash
$ hlx --version
5.7.0
```

**Additional context**
As a workaround you can use `hlx up --no-local-repo` instead of `hlx up`

stefan-guggisberg:
regression of #913

stefan-guggisberg:
closed via #1201

Issue #913, opened 2019-05-24
--local-repo=. should be default


tripodsan:
**Description**
once deployed, the content repo is set to the git origin and the local development uses the remote repository, which is not intuitive.

## suggest

make `--local-repo` the default. if devs want to test the remote content repository, they can specify `--no-local-repo`

kptdobe:
Related to #711

stefan-guggisberg:
Reopening, the fix caused a regression (#1189)and had to be reverted. Issue still exists.

tripodsan:
as I mentioned somewhere else, the initial idea of this issue was, that for local development, it should force a local repo, if not specified otherwise.

- new repo, no origin -> simulator takes care. don't specify `--local-repo=.`
- repo with origin -> default to `--local-repo=.`
- if any other `--local-repo` or `--no-local-repo` is specified, use as specified.

PR #1201, opened 2019-10-31
fix(up.cmd): support `hlx up` in local repo checkout without configured origin

stefan-guggisberg:
## Related Issues

#1189

### DIFF
--- a/src/up.cmd.js
+++ b/src/up.cmd.js
@@ -159,20 +159,24 @@ class UpCommand extends BuildCommand {
     }
 
     // check all local repos
-    const localRepos = await Promise.all(this._localRepos.map(async (repo) => {
+    const localRepos = (await Promise.all(this._localRepos.map(async (repo) => {
       const repoPath = path.resolve(this.directory, repo);
       if (!await fse.pathExists(path.join(repoPath, '.git'))) {
         throw Error(`Specified --local-repo = ${repo} is not a git repository.`);
       }
       const gitUrl = await GitUtils.getOriginURL(repoPath);
       if (!gitUrl) {
+        if (repo === '.') {
+          // local git repo without remote, skip
+          return null;
+        }
         throw Error(`Unable to determine origin url for --local-repo = ${repo}`);
       }
       return {
         gitUrl,
         repoPath,
       };
-    }));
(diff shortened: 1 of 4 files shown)


test/testUpCmd.js
```diff
@@ -27,6 +27,7 @@ const {
   createTestRoot,
 } = require('./utils.js');
 
+const DemoCommand = require('../src/demo.cmd');
```
tripodsan: is this really required?

src/up.cmd.js
```diff
@@ -159,20 +159,24 @@ class UpCommand extends BuildCommand {
...
     // check all local repos
-    const localRepos = await Promise.all(this._localRepos.map(async (repo) => {
+    const localRepos = (await Promise.all(this._localRepos.map(async (repo) => {
       const repoPath = path.resolve(this.directory, repo);
       if (!await fse.pathExists(path.join(repoPath, '.git'))) {
         throw Error(`Specified --local-repo = ${repo} is not a git repository.`);
       }
       const gitUrl = await GitUtils.getOriginURL(repoPath);
       if (!gitUrl) {
+        if (repo === '.') {
```
tripodsan: this fixes it, but I think the error is a different one, such as the `.` is a new default.
this was changed #913. I think the check for a local repo should be different.

the original idea was that you can specify a local repo that overrides the resolution for git repositories specified in the strains. Before any strains are setup and before the local git has no origin, then the simulator anyways uses the cwd.

so I think that #913 should be fixed in a way, that it sets `.` as default, only if the local repo has an origin.

rofe:
Should we also update `testDemoUp.js` to use `.` as local repo like it would on the command line?

stefan-guggisberg:
> Should we also update testDemoUp.js to use . as local repo like it would on the command line?

The `--locol-repo .` default is applied conditionally. `testDemoUp.js` is using lower-level components (`DemoCommand`, `UpCommand`) and does not cover command line parameter handling and defaults. For a more realistic test case `testDemoUp.js` should be rewritten using the `CLI`, similar to https://github.com/adobe/helix-cli/blob/d0d73880f13e0aba4444ab6311143822377dd5e4/test/testUpCli.js#L171-L177

Issue #1205, opened 2019-11-04
Tweak renovatebot config?


trieloff:
I'd like to try a couple of renovatebot config changes in this repo (and roll them out to all repos once we have confidence)

1. rename the "any" group to "external" – this just makes things more clear
2. make sure that internal dependency changes do not get caught up in "update any" PRs like https://github.com/adobe/helix-cli/pull/1196 – this way we can keep moving on internal changes even if an external changes blocks the merge
3. enforce semantic commits (some repos like Helix Pipeline still don't get semantic commits)
4. enable automerge for patch updates (this way we don't have to review these minor changes anymore)
5. enable automerge for internal minor and patch releases

PR #1212, opened 2019-11-06
Renovate renovatebot config

trieloff:
Fix for #1205


trieloff:
Let's see how this goes.

Issue #1242, opened 2019-12-12
Tests should accept a specific pipeline version (git repo) as a parameter


kptdobe:
In order to test the current cli master code with a specific pipeline version (for example when developing a new feature in the pipeline), the test needs to accept a pipeline version (for the smoke tests, we only need a git branch / git url with a branch) as a parameter.

tripodsan:
this is fixed.

PR #1243, opened 2019-12-12
Tests should accept a specific pipeline version (git repo) as a parameter

kptdobe:
PR for #1242 

With this you can run: 

```
env HLX_CUSTOM_PIPELINE="https://github.com/adobe/helix-pipeline.git#winston-compatability" npm run test
```

to run the tests with a custom pipeline.

This introduces a HLX_CUSTOM_PIPELINE env variable (or customPipeline) which would be the npm dependency to install - potential value could be `@adobe/helix-pipeline`, `@adobe/helix-pipeline@<version>` or `https://github.com/adobe/helix-pipeline.git#branch` (anything npm install accepts).

This also fixes the env variables interferences between tests (env variables get lost after first call of `clearHelixEnv` and are not available to following tests).


src/build.js
```diff
@@ -28,6 +28,14 @@ module.exports = function build() {
     builder: (yargs) => {
       yargsBuild(yargs);
```
tripodsan: why not add to `yargsBuild` ? it is needed also in all other commands, like `up` and `package` and `deploy` ...

src/builder/ModuleHelper.js
```diff
@@ -130,7 +130,7 @@ class ModuleHelper {
     }
   }
 
-  async installModule(name) {
+  async installModule(name, installer) {
```
tripodsan: I think the name `installer` is very confusing. what about `descriptor` ?

kptdobe: I knew you will find a better name for it. I could not :)

src/build.js
kptdobe: I saw that `up` extends `build` thus I thought I do not it here.
`deploy` does not seem to run a `build` thus we would not need it here.
`package` needs it, correct, I miss it.
I'll add the arg to `yargsBuild`.

kptdobe: ok, found it. `deploy` calls `package` which calls `build`...

kptdobe:
@tripodsan I added support for the option in `build`, `up`, `package` and `deploy`, together with corresponding tests.
I do not understand though the codecov drop...

kptdobe:
> I think we might revisit the way we collect common argument...it's kind of stupid to add the same stuff to up,build,deploy,package....

I agree, the parameter handling could be generalised.

PR #1275, opened 2020-01-22
chore(deps): update dependencies

tripodsan:


tripodsan:
> Looks good, but why do we have to do this manually?

I don't know... but for example fastly-native-promises was 3 versions back and tidelift complained about a transient dep.

PR #1279, opened 2020-01-24
Publish queries

trieloff:
Adds support for https://github.com/adobe/helix-publish/pull/288

### DIFF
--- a/src/abstract.cmd.js
+++ b/src/abstract.cmd.js
@@ -14,7 +14,7 @@
 
 const EventEmitter = require('events');
 const chalk = require('chalk');
-const { HelixConfig } = require('@adobe/helix-shared');
+const { HelixConfig, IndexConfig } = require('@adobe/helix-shared');
 const { getOrCreateLogger } = require('./log-common');
 const ConfigUtils = require('./config/config-utils.js');
 
@@ -24,10 +24,12 @@ class AbstractCommand extends EventEmitter {
     this._initialized = false;
     this._logger = logger || getOrCreateLogger();
     this._helixConfig = new HelixConfig().withLogger(this._logger);
+    this._indexConfig = new IndexConfig().withLogger(this._logger);
   }
 
   withDirectory(dir) {
     this._helixConfig.withDirectory(dir);
+    this._indexConfig.withDirectory(dir);
     return this;
   }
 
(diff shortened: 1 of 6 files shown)


trieloff:
`helix-index.yaml` is read and passed on to `helix-publish`, so that https://github.com/adobe/helix-publish/pull/288 can create proxies for each index/query defined in `helix-index.yaml`.

tripodsan:
> `helix-index.yaml` is read and passed on to `helix-publish`, so that [adobe/helix-publish#288](https://github.com/adobe/helix-publish/pull/288) can create proxies for each index/query defined in `helix-index.yaml`.

ah, ok. but on the long run, this will not work. since the `helix-index.yaml` is defined on a project repository, that might not have its own fastly service.

e.g. _theblog_ defines the helix-index.yaml, but we should not deploy this to helix-pages. and we should also not create an own fastly-service for _theblog_.

so, the helix-pages VCL needs to contain special cases for _all_ projects that use an index...
or we need to start creating a fastly-service for _theblog_ that handles the queries and proxies the rest to helix-pages.

trieloff:
I don't see this as a Helix problem per se (and therefore don't agree with removing the feature from `helix-publish` or blocking the PR in `helix-cli`), but as a Helix Pages problem.

What we can do in Helix Pages is define a couple generic queries and indexes that can be shared by consumers. Helix Pages users can override how these indexes are populated, but not how they are queried.

At a later point in time, we can also add a Runtime-backed query type that will be less efficient in execution than the straight Fastly-Algolia proxy, but I'd start with what we have and improve incrementally.

tripodsan:
> Helix Pages users can override how these indexes are populated, but not how they are queried.

In any case, we need to treat those special. i.e. the name of the index depends on the repository, so for each customer, this name needs to be either passed as url parameter, or it needs to be addressed by a separate service config.

even with this change, since the `helix-index.yaml` is not in the helix-pages repository, but the `theblog` repository, this would not help for theblog. of course it ca be copied there, but this is kind of weird...

I think we should avoid deviating too much from a good solution, just to get _theblog_ to work..

I think a better solution would be to:
- create a fastly service just for theblog (I think we already have one)
- add the special query stuff there
- use the helix-bot to trigger update and cacheflush etc...

trieloff:
> In any case, we need to treat those special. i.e. the name of the index depends on the repository, so for each customer, this name needs to be either passed as url parameter, or it needs to be addressed by a separate service config.

No, the `helix.vcl` uses the `owner` and `repo` for the index name, and that is handled well in the Helix Pages extensions.

I'll provide a suggestion for a generic fallback later, but I'd be glad if we could merge the PR as is

PR #1362, opened 2020-04-13
feat(cli): better default parameter parsing

### LINKED ISSUE #1352: Better default parameter support (env, json, files)
similar to #1266, the `--default` parameter should also be usable as env variable.

tripodsan:
fixes #1352

### DIFF
--- a/src/yargs-params.js
+++ b/src/yargs-params.js
@@ -9,39 +9,84 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
+const path = require('path');
+const fse = require('fs-extra');
+const dotenv = require('dotenv');
+
+/**
+ * Decoded the params string. First as JSON and if this fails, as ENV format.
+ *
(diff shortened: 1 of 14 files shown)


src/yargs-params.js
```diff
@@ -9,39 +9,86 @@
...
+  let content = params.trim();
+  if (cwd) {
+    const filename = path.resolve(cwd, content);
+    if (!fse.existsSync(filename)) {
+      throw Error(`Specified param file does not exist: ${path.relative(process.cwd(), filename)}`);
+    }
+    content = fse.readFileSync(filename, 'utf-8').trim();
+  }
+  let data;
+  if (typeof params === 'object') {
```
kptdobe, line 33: How can `params` be an object here ? If it is an object, the previous `params.trim()` would most likely fail, no ?
Unit test those functions would reveal the flaws...

tripodsan: good catch....copy/paste leftover from another project ;-)

PR #1412, opened 2020-05-29
test(up-cmd): adjust tests for new helix pipeline version

ramboz:
With the new pipeline changing how we order the meta types for better consistency, we need to update the cli's up command tests to match the new order

fix adobe/helix-pipeline#issues/669

### DIFF
--- a/test/specs/json_response.json
+++ b/test/specs/json_response.json
@@ -13,8 +13,8 @@
             "nb-blockquote-1",
             "nb-delete-1",
             "nb-heading-1",
-            "is-text-html-blockquote",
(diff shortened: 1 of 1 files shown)


ramboz:
@trieloff What's the best-practice here to get this in with https://github.com/adobe/helix-pipeline/pull/728?

The adjusted test won't pass without the new pipeline, but the new pipeline PR fails as long as the test is not in… Is there a clean way to do this?

trieloff:
I don't know. Just merge as admin? Maybe @kptdobe has a cleaner solution.

PR #1413, opened 2020-06-01
Add timestamp

### LINKED ISSUE #1409: add timestamp and project version to package during deploy
when helix-cli creates (or updates) a new openwhisk package, it should record the project version and timestamp.

for example for helix-pages it would be nice to know, which version a specific package belongs to and when it was created.

tripodsan:
fixes #1409

### DIFF
--- a/src/deploy.cmd.js
+++ b/src/deploy.cmd.js
@@ -62,6 +62,8 @@ class DeployCommand extends AbstractCommand {
     this._epsagonToken = null;
     this._coralogixAppName = null;
     this._coralogixToken = null;
+    this._updated = null;
+    this._updatedBy = null;
   }
 
   get requireConfigFile() {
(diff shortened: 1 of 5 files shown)


src/deploy.cmd.js
```diff
@@ -209,6 +222,8 @@ class DeployCommand extends AbstractCommand {
     } catch (e) {
       this._pkgJson = {};
     }
+    this._updated = this._updated || new Date().getTime();
+    this._updatedBy = this._updatedBy || os.userInfo().username;
```
trieloff, line 226: I’d like us to find a more privacy-friendly default for that. Having your user name in an annotation feels sneaky and surprising, especially when it’s done without an opt-in.

trieloff:
We could create alias names: build a list of first and last names, then randomly pick one of each. Initialize the random seed with either the OS username or the name configured in Git (probably more reliable anyway). This makes the deploying person still identifiable, but only with their consent.

> According to ‘hlx name’ I am deploying as “Sally Fish”.

src/deploy.cmd.js
tripodsan: hmm..ok. I remove the updatedBy for now.

PR #1456, opened 2020-07-01
feat(request): request is deprecated; replace with helix-fetch

### LINKED ISSUE #1455: request is deprecated; replace with helix-fetch
[request](https://github.com/request/request) is deprecated and should be replaced.
best candidate is [helix-fetch](https://github.com/adobe/helix-fetch)

kptdobe:
Fix #1455

### DIFF
--- a/test/specs/builder/src/helpx_html.pre.js
+++ b/test/specs/builder/src/helpx_html.pre.js
@@ -13,7 +13,7 @@
 /* eslint-disable no-param-reassign */
 
 const _ = require('lodash/fp');
-const request = require('request-promise-native');
+const { fetch } = require('@adobe/helix-fetch');
 
 function collectMetadata(req, logger) {
   const options = {
@@ -28,8 +28,14 @@ function collectMetadata(req, logger) {
 
   logger.debug(`Fetching Git Metadata from ${options.uri}`);
 
-  return request(options)
-    .then((metadata) => {
+  return fetch(options.uri, options)
+    .then(async (result) => {
+      if (!result.ok) {
+        const e = new Error(`${result.status} - "${await result.text()}"`);
+        e.statusCode = result.status;
+        throw e;
+      }
+      const metadata = await result.json();
       logger.debug('Got git metadata');
       return metadata;
     })
--- a/test/specs/builder/src/helpx_html.pre.js
+++ b/test/specs/builder/src/helpx_html.pre.js
@@ -29,13 +29,13 @@ function collectMetadata(req, logger) {
   logger.debug(`Fetching Git Metadata from ${options.uri}`);
 
   return fetch(options.uri, options)
-    .then(async (result) => {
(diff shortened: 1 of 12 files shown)


kptdobe:
Tests are failing because `nock` does not intercept the `helix-fetch` http2 POST requests.

stefan-guggisberg:
> Tests are failing because `nock` does not intercept the `helix-fetch` http2 POST requests.

yup, known problem. neither `nock` nor `pollyjs` can intercept `HTTP2` requests. you have to enforce `http1` for the tests.

kptdobe:
I do not understand what happens with the codecov. While it should decrease a little bit (you need more code with `helix-fetch` to do the things than with `request`), I do not get why it completely drops...

kptdobe:
@tripodsan I need your help to fix the tests: I replace `rp` by `helix-fetch` in the test code here https://github.com/adobe/helix-cli/pull/1456/files#diff-e5b829b91f089a4b948a5e61343bb0d1R87 but the test hangs for 60s and then times out. But I do not understand why, especially what is this request supposed to "unblock".

kptdobe:
I think I found it: I added @stefan-guggisberg 's trick, use `context.disconnectAll()` and it seems to work.

But this is really weird because I consume the body and it does not seem to be enough: the server does not stop and it is because there is still something connected...

test/specs/builder/src/helpx_html.pre.js
```diff
@@ -28,8 +28,14 @@ function collectMetadata(req, logger) {
 
   logger.debug(`Fetching Git Metadata from ${options.uri}`);
 
-  return request(options)
-    .then((metadata) => {
+  return fetch(options.uri, options)
+    .then(async (result) => {
```
stefan-guggisberg, line 32: nitpicking... `fetch` returns a `Response` object. `resp` instead of `result` would make it more obvious.

stefan-guggisberg:
> But this is really weird because I consume the body and it does not seem to be enough: the server does not stop and it is because there is still something connected...

could be due to keep-alive. would need to investigate though.

test/testAuthCmd.js
```diff
@@ -151,4 +159,40 @@ describe('Integration test for auth', () => {
...
+        await res.buffer();
+        // required so that the server can stop properly
+        fetchAPI.disconnectAll();
+      });
+
+    try {
+      await cmd.run();
+      assert.fail('auth should fail');
+    } catch (e) {
+      assert.equal(e.message, '500 - "GitHub API has an issue"');
```
tripodsan, line 193: this fails under windows: https://circleci.com/gh/adobe/helix-cli/13307?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

kptdobe: Fixed! I had removed too much ;)

PR #1501, opened 2020-08-11
fix(deploy): pin major version for helix-static

### LINKED ISSUE #1498: Change Static Action from hlx--static to static@v1
The hlx--static action links to static@latest, which has potentially breaking changes.

trieloff:
fixes #1498

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/deploy.cmd.js
+++ b/src/deploy.cmd.js
@@ -576,14 +576,14 @@ Alternatively you can auto-add one using the {grey --add <name>} option.`);
     let staticactionname = '/hlx--static';
     if (!this._dryRun) {
       // probe Helix Static action for version number
-      await fetch('https://adobeioruntime.net/api/v1/web/helix/helix-services/static@latest').then(async (res) => {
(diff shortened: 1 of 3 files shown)


tripodsan:
hmm...something might be broken. we used to call the `static@latest` in order to retrieve it's true version (x-version header). and then use this for linking the `hlx--static`. this was to ensure that a once deployed action package remains stable, even if we change the helix-static sequence.

is this still the case?

however, I was thinking this over again - we have so many services that are crucial and the customer has no control over (dispatch, content proxy, redirect, ...) so why should `static` be special? maybe we should remove the `helix--static` completely and directly call `helix-services/static@v1` in dispatch....

trieloff:
> is this still the case?

The check is still there, but it seems to fail in practice, so that the fallback is always used. 

I don't think this special handling of `static` is worth it and would rather invoke static directly from dispatch.

tripodsan:
> I don't think this special handling of `static` is worth it and would rather invoke static directly from dispatch.

let's do this!

PR #1580, opened 2020-12-06
fix(publish): enforce http/1 to avoid problems with ethos

tripodsan:
The upcoming/new I/O Runtime environment is Azure-based and supports HTTP/2. This might lead to issues with long-lived HTTP/2 sessions.


stefan-guggisberg (commented):
I don't think this is necessary. The connection established by the CLI is short-lived and should therefore not be a problem.

trieloff (approved):
I don't think this is necessary.

tripodsan:
> I don't think this is necessary. The connection established by the CLI is short-lived and should therefore not be a problem.

hmmm...it failed for me a couple of times....maybe it was another error.

PR #1624, opened 2021-01-27
feat(deps): upgrade to helix-fetch v2

tripodsan:

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "@adobe/fastly-native-promises": "^1.19.2",
-    "@adobe/helix-fetch": "^1.9.2",
+    "@adobe/helix-fetch": "2.0.1",
     "@adobe/helix-log": "4.5.3",
     "@adobe/helix-shared": "7.19.2",
     "@adobe/helix-simulator": "5.1.20",
@@ -41,7 +41,7 @@
     "archiver": "^5.1.0",
     "camelcase": "^6.2.0",
     "chalk": "4.1.0",
-    "chokidar": "^3.4.3",
+    "chokidar": "3.4.3",
     "dotenv": "^8.0.0",
     "ferrum": "^1.7.0",
     "fs-extra": "^9.0.1",
--- a/src/deploy.cmd.js
+++ b/src/deploy.cmd.js
@@ -12,7 +12,7 @@
 
 'use strict';
 
-const { fetch } = require('@adobe/helix-fetch');
+const fetchAPI = require('@adobe/helix-fetch');
 const chalk = require('chalk');
 const ow = require('openwhisk');
 const glob = require('glob');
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
     "archiver": "^5.1.0",
     "camelcase": "^6.2.0",
     "chalk": "4.1.0",
-    "chokidar": "^3.4.3",
+    "chokidar": "3.4.3",
```
tripodsan, line 44: this was needed, because I got errors on macOS.
see: https://github.com/fsevents/fsevents/issues/355

src/deploy.cmd.js
```diff
@@ -271,7 +278,6 @@ class DeployCommand extends AbstractCommand {
 
     const followoptions = {
       method: 'POST',
-      json: true,
```
tripodsan, line 274: I think the `json` option doesn't have a meaning in helix-fetch, right?

trieloff: It did not in 1.0

tripodsan:
now the windows tests fail....oh my..... I'll look at them tomorrow...but feel free :-)

src/deploy.cmd.js
stefan-guggisberg: It did. You could specify the JSON payload (plain JS object) of `POST`, `PUT`, `PATCH` requests via the `json` option. 

In v2 the `json` option is not supported anymore but you can do the same via the `body` option:
https://github.com/adobe/helix-fetch/blob/main/v2-UPGRADE-GUIDE.md#post-json

Primitive `body` values however (e.g. `true`, `42`) are not (yet) interpreted as JSON payload. I guess the `json: true` line here is a leftover from the time when we used the `request` module.

src/deploy.cmd.js
```diff
@@ -271,7 +278,6 @@ class DeployCommand extends AbstractCommand {
...
       auth,
```
stefan-guggisberg, line 281: Note that this is a `request` specific option. It was never supported by `helix-fetch` (nor by `node-fetch`) ... 

This code never worked.

src/deploy.cmd.js
```diff
@@ -331,7 +337,6 @@ class DeployCommand extends AbstractCommand {
 
       const triggeroptions = {
         method: 'POST',
-        json: true,
         auth,
```
stefan-guggisberg, line 340: Note that this is a `request` specific option. It was never supported by `helix-fetch` (nor by `node-fetch`) ... 

This code never worked.

stefan-guggisberg (commented):
LGTM, apart from the `auth` option which obviously didn't work for a long time and nobody noticed.

stefan-guggisberg:
related: https://github.com/adobe/helix-home/issues/182

tripodsan:
> now the windows tests fail....oh my..... I'll look at them tomorrow...but feel free :-)

hmmm:

```
  2) hlx package (custom pipeline)
       package installs the correct custom pipeline via directory:
     Error: exec: fatal: cannot create directory at 'test/fixtures/Testing-HTML-Pipeline-universal-_1522650794/html-pipe-complains-with-a-specific-message-for-mdast-nodes-wih-extra-properties-when-con_3402660966': Filename too long
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'
```

tripodsan:
the  windows tests on circleci are different than the ones I get locally...

Issue #1794, opened 2021-06-29
"hlx up" fails with "Could not find HEAD." on repos that are submodules


maxakuru:
**Description**
I pulled a repo as a submodule to keep all helix repos together to explore things in a single place, but when running `hlx up` on a repo that doesn't have a `.git` folder (instead, a `.git` file that points to the parent's gitdir) the command fails with the error: `Could not find HEAD.`

It looks like this is due to the way the gitdir is resolved, currently the UpCommand assumes it's in a regular repository. GitUtils also seem to use the default gitdir behavior for all APIs (which ends up being dir/ + .git).

**To Reproduce**
1. `git init`
2. `git submodule add <some helix repo url>`
3. `cd repo`
4. `hlx up`

**Expected behavior**
Repos as a submodules is a popular format (but maybe uncommon for helix sites). I think this should be supported, or if not, then the `.git` path should be checked for "directory-ness" and not just existence, so the error message is more clear.

**Version:**
`13.4.49`

I have a fix working and could add some unit tests to PR this, but wanted to find out if this has been discussed already.. I noticed issue #614 speaks to submodules inside the project, but not the other way around.

tripodsan:
my personal opinion is, that submodules cause more problems than they solve :-)
however, I think it is good to support all kind of setups.

stefan-guggisberg:
I second @tripodsan's statement, having used submodules in some projects in the past. I haven't seen projects using submodules in recent years and I'd question that they're a popular choice.

Unless there's a real use case IMHO I don't think it's worth supporting it.

maxakuru:
@tripodsan @stefan-guggisberg I agree submodules aren't great and monorepos in general are polarizing. It's hard to argue against the popularity of monorepos for dev environments, though, whether using submodules or another tool. It may be rare, but if I hit this error and wasn't motivated to find out why, I might have lost interest. At very least a better error message makes sense, imo

I didn't realize the potential impact to support it in other components. If the consensus is "not needed" or "not wanted" I'll gladly drop the PR or convert it to an error message PR... wdyt

stefan-guggisberg:
An error msg like e.g. „git submodules aren‘t supported“ would be good.

PR #1796, opened 2021-07-01
fix(cli): add error message for hlx up command in submodule repo

maxakuru:
## Related Issues
#1794

### DIFF
--- a/src/up.cmd.js
+++ b/src/up.cmd.js
@@ -180,9 +180,18 @@ class UpCommand extends BuildCommand {
   async setup() {
     await super.init();
     // check for git repository
-    if (!await fse.pathExists(path.join(this.directory, '.git'))) {
-      throw Error('hlx up needs local git repository.');
+    try {
+      const stat = await fse.lstat(path.join(this.directory, '.git'));
+      if (stat.isFile()) {
+        throw Error('git submodules are not supported.');
+      }
+    } catch (e) {
+      if (e.code === 'ENOENT') {
+        throw Error('hlx up needs local git repository.');
+      }
+      throw e;
     }
+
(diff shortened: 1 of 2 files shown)


src/up.cmd.js
```diff
@@ -180,9 +180,18 @@ class UpCommand extends BuildCommand {
   async setup() {
     await super.init();
     // check for git repository
-    if (!await fse.pathExists(path.join(this.directory, '.git'))) {
-      throw Error('hlx up needs local git repository.');
+    try {
+      const stat = await fse.lstat(path.join(this.directory, '.git'));
```
tripodsan, line 184:
```suggestion
      const stat = await fse.lstat(path.resolve(this.directory, '.git'));
```

tripodsan (dismissed):
I would use `path.resolve()`. otherwise, LGTM.

maxakuru:
@tripodsan what's the process for merges? squash? let approver merge?

tripodsan:
> @tripodsan what's the process for merges? squash? let approver merge?

no, it's usually the PR author that does the merge. if squash or merge is a personal choice :-) but all commits need to adhere to the commit message rules (https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular)

so if you have a lot for small (`wip`) commits, you should either squash them before the merge, or do a sqash commit in github. but if you have a _nice_ set of commits, feel free to just merge them.

note, that semantic-release will create a released base on the commit types. so if you have a `fix: ...` commit in the history, if will create a release, of all the commits are only `chore:` it will not create a release.

tripodsan (approved):

PR #1888, opened 2021-11-08
Local head html support

tripodsan:


src/server/HeadHtmlSupport.js
```diff
@@ -0,0 +1,107 @@
...
+      this.localHtml = this.sanitize(await fs.readFile(this.filePath, 'utf-8'));
+      this.localStatus = 200;
+      this.log.debug('loaded local head.html from from', this.filePath);
+    } catch (e) {
+      this.log.error(`error while loading local head.html from ${this.filePath}: ${e.code}`);
+      this.localStatus = 404;
+    }
+  }
+
+  // eslint-disable-next-line class-methods-use-this
```
kptdobe, line 57: Why not putting `sanitize` in `utils` or at least `static` ?

kptdobe (approved):

src/server/HeadHtmlSupport.js
tripodsan: it's nothing we use so often... and initially I though, we should log potential parsing errors (`this.log`)....but there are none.

PR #1904, opened 2021-12-02
feat: add --print-index to help debug query.yaml

### LINKED ISSUE #1903: print indexed values while surfing
configuring the `helix-query.yaml` is difficult as you don't see which values are extracted until they are added to the index.

it would be useful to enable a mode (`hlx up --print-indexed`) which causes the cli to print all indexed properties/value when surfing during local dev. it should work with live-reload, so that the cli doesn't need to be restarted when editing the `helix-query.yaml`

tripodsan:
fixes #1903

example output:

image

### DIFF
--- a/src/up.cmd.js
+++ b/src/up.cmd.js
@@ -26,6 +26,7 @@ export default class UpCommand extends AbstractCommand {
     this._open = '/';
     this._liveReload = false;
     this._pagesUrl = null;
+    this._printIndex = false;
   }
 
   withHttpPort(p) {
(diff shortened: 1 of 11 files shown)


src/up.cmd.js
```diff
@@ -48,15 +49,16 @@ export default class UpCommand extends AbstractCommand {
...
+
   get project() {
     return this._project;
   }
 
   async stop() {
-    if (this._watcher) {
-      await this._watcher.close();
-      this._watcher = null;
-    }
```
tripodsan, line 59: left over from previous refactor.

dominique-pfister (approved):
Awesome 🎉

rofe (approved):
Cool!

PR #1952, opened 2022-03-01
Implement kill switch

### LINKED ISSUE #1939: Add kill URL to shut down the server?
When working on multiple helix projects at the same time, you will often find out that `hlx up` fails, because port 3000 is already in use. What if, instead of searching for the console tab that has the running process, you could just `curl http://localhost/.kill && hlx up`, which would shut down the existing process, free the port and allow starting the new server without leaving the console?

trieloff:
- feat(server): add `/.kill` switch to quickly turn the server off
- feat(cli): add `--kill` option (default: on) to kill other Helix CLI server instances on the same port

fixes #1939

### DIFF
--- a/src/server/HelixServer.js
+++ b/src/server/HelixServer.js
@@ -133,7 +133,7 @@ export default class HelixServer extends EventEmitter {
   async start() {
     const { log } = this;
     if (this._port !== 0) {
-      if (this._kill) {
+      if (this._kill && await utils.checkPortInUse(this._port)) {
         await fetch(`http://localhost:${this._port}/.kill`);
       }
       const inUse = await utils.checkPortInUse(this._port);
--- a/src/server/HelixServer.js
+++ b/src/server/HelixServer.js
@@ -107,6 +107,10 @@ export default class HelixServer extends EventEmitter {
 
   async setupApp() {
     const handler = asyncHandler(this.handleProxyModeRequest.bind(this));
+    this._app.get('/.kill', (req, res) => {
+      res.send('Goodbye!');
+      process.exit();
+    });
     this._app.get('*', handler);
     this._app.post('*', handler);
   }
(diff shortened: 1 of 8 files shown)


src/server/HelixServer.js
```diff
@@ -127,6 +133,9 @@ export default class HelixServer extends EventEmitter {
   async start() {
     const { log } = this;
     if (this._port !== 0) {
+      if (this._kill) {
+        await fetch(`http://localhost:${this._port}/.kill`);
+      }
```
tripodsan, line 138: I would not do this pre-emptively, but rather after checking `inUse`. and you might need to wait a little for the other process to stop and release the socket.

src/up.js
```diff
@@ -51,6 +51,11 @@ export default function up() {
           type: 'int',
           default: 3000,
         })
+        .option('kill', {
+          describe: 'Kill existing Helix CLI running on the above port',
+          type: 'boolean',
+          default: true,
+        })
```
tripodsan, line 58: I would prefer a less negative term, like `--force`

test/server.test.js
```diff
@@ -50,10 +50,11 @@ describe('Helix Server', () => {
 
       const project2 = new HelixProject()
         .withCwd(cwd)
+        .withKill(false)
         .withHttpPort(project.server.port);
       await project2.init();
       try {
-        await project.start();
+        await project2.start();
```
tripodsan, line 57: good catch :-)

test/server.test.js
```diff
@@ -77,6 +78,20 @@ describe('Helix Server', () => {
...
+    const cwd = await setupProject(path.join(__rootdir, 'test', 'fixtures', 'project'), testRoot);
+    const project = new HelixProject()
+      .withCwd(cwd)
+      .withHttpPort(0);
+    await project.init();
+    try {
+      await project.start();
+      await assertHttp(`http://localhost:${project.server.port}/.kill`, 200, 'expected_goodbye.txt');
+    } finally {
+      await project.stop();
```
tripodsan, line 91: maybe check `project.stopped` or something....

tripodsan (approved):

PR #1956, opened 2022-03-11
feat: filesystem cache for hlx up

kptdobe:
During my importer explorations, I wrote a simple filesystem cache for `hlx up`.

Usage `hlx up --cache .cache/` for example.

This will store all requested resources in the `.cache` folder: one file for the body and one json file with { headers, status } of the request.

Using `hlx up --cache .cache/ --pagesUrl https://www.customer.com` is a super convient way to get a proxy to a customer website (no cors, no csp) and the cache allows to save all the content... Running imports in then super straight forward and fast (could be even faster if the local sever would implement http2).

I think this is twisting `hlx up` nature and is not the right flow. But before finding where all of this must go and the correct import journey (create another cli, full chrome extension...?), I would like to get this in hlx cli to be able to explore more. I can still remove it later when not needed.

### DIFF
--- a/src/server/utils.js
+++ b/src/server/utils.js
@@ -110,7 +110,7 @@ const utils = {
         fileName = `${fileName}.${qs}`;
       }
     }
-    const filePath = path.resolve(directory, fileName);
+    const filePath = path.join(directory, fileName);
     return filePath;
   },
 
--- a/src/server/utils.js
+++ b/src/server/utils.js
@@ -110,7 +110,7 @@ const utils = {
         fileName = `${fileName}.${qs}`;
       }
     }
-    const filePath = path.join(directory, fileName);
+    const filePath = path.resolve(directory, fileName);
     return filePath;
   },
 
--- a/src/server/utils.js
+++ b/src/server/utils.js
@@ -92,6 +92,13 @@ const utils = {
     return body;
   },
 
+  /**
+   * Computes a path to store a cache objet
+   * @param {string} pathname the url pathname
+   * @param {string} qs the url query string
+   * @param {string} directory a local directory path
+   * @returns {string} the computed path
+   */
   computePathForCache(pathname, qs, directory) {
     let fileName = pathname;
     if (qs) {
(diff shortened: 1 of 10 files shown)


src/up.js
```diff
@@ -68,7 +68,11 @@ export default function up() {
           describe: 'The origin url to fetch pages content from.',
           type: 'string',
         })
-        .group(['pages-url', 'livereload', 'no-livereload', 'open', 'no-open', 'print-index'], 'Helix Pages Options')
+        .option('cache', {
+          describe: 'Path to local folder to cache the responses',
```
trieloff, line 72:
```suggestion
        .option('alpha-cache', {
          describe: 'Path to local folder to cache the responses (note: this is an alpha feature, it may be removed without notice',
```

src/up.js
```diff
@@ -87,6 +91,7 @@ export default function up() {
         .withPagesUrl(argv.pagesUrl)
         .withPrintIndex(argv.printIndex)
         .withKill(argv.stopOther)
+        .withCache(argv.cache)
```
trieloff, line 94:
```suggestion
        .withCache(argv.alphaCache)
```

trieloff (commented):
I'm fine with adding this as a temporary feature, but I'd mark it in a way so that we don't need a new major version when we remove it.

src/server/utils.js
```diff
@@ -91,6 +92,50 @@ const utils = {
...
+    if (qs) {
+      const index = fileName.lastIndexOf('.');
+      if (index > -1) {
+        // inject qs as b64 in filename before extension
+        fileName = `${fileName.substring(0, index)}.${qs}${fileName.substring(index)}`;
+      } else {
+        fileName = `${fileName}.${qs}`;
+      }
+    }
+    const filePath = path.join(directory, fileName);
```
tripodsan, line 106:
```suggestion
    const filePath = path.resolve(directory, fileName);
```

src/server/utils.js
```diff
@@ -91,6 +92,50 @@ const utils = {
...
+    return filePath;
+  },
+
+  async writeToCache(pathname, qs, directory, { body, headers, status }, logger) {
```
tripodsan, line 110: jsdoc?

tripodsan (approved):

trieloff (dismissed):

PR #1993, opened 2022-04-25
feat: import - support 301

kptdobe:
The 301 support got lost while implementing the `hlx import` command. Re-adding it.

### DIFF
--- a/src/server/HelixImportServer.js
+++ b/src/server/HelixImportServer.js
@@ -136,8 +136,14 @@ export default class HelixServer extends EventEmitter {
     const respHeaders = ret.headers.plain();
     delete respHeaders['content-encoding'];
     delete respHeaders['content-length'];
-    delete respHeaders.location;
     respHeaders['access-control-allow-origin'] = '*';
+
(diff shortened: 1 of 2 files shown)


src/server/HelixImportServer.js
```diff
@@ -136,9 +136,15 @@ export default class HelixServer extends EventEmitter {
     const respHeaders = ret.headers.plain();
     delete respHeaders['content-encoding'];
     delete respHeaders['content-length'];
-    delete respHeaders.location;
     respHeaders['access-control-allow-origin'] = '*';
 
+    if (respHeaders.location && !respHeaders.location.startsWith('/')) {
```
tripodsan, line 141: don't you need to check for 301 or 302? otherwise a `200` with a `location` header will also rewrite.

kptdobe: from the spec, the `location` header is only meaningful for 201 and 3xx.
In any case, it only makes the url relative which would be useful in any case (to avoid to have to rewrite the origin and use the `host` query param. It cannot hurt I think.

tripodsan (approved):

trieloff (approved):

PR #2033, opened 2022-06-24
feat: install and build importer-ui

### LINKED ISSUE #2031: [import] git clone the import-tools repo
When running `hlx import`, the cli opens a browser page to the helix-webui tool. But this requires that end-user has manually git cloned the repo in the `tools/importer` folder. The git clone could be automated:

- when `hlx import` starts, check if `tools/importer/helix-importer-tools` (or whatever the final project name is, see #2030) exists.
- if does not exist: `git clone` the repo and run `npm i && npm run build` (see latest README instructions for the exact sequence)
- if it does exist: run `git pull` and `npm i && npm run build` to get and build the latest version.

kptdobe:
Fix #2031 

The importer-ui project build process is really slow (more than one minute), I will look at that separately. 
Another approach could also be to package the importer UI together with the hlx cli...

I also did not implement the git fetch + build if the project is already there mainly because the build process is super slow. Impact is that `hlx import` would always be slow to start.

### DIFF
--- a/src/import.cmd.js
+++ b/src/import.cmd.js
@@ -15,12 +15,10 @@ import opn from 'open';
 import chalk from 'chalk-template';
 import git from 'isomorphic-git';
 import http from 'isomorphic-git/http/node/index.js';
-import shell from 'shelljs';
 import HelixImportProject from './server/HelixImportProject.js';
 import pkgJson from './package.cjs';
 import AbstractCommand from './abstract.cmd.js';
 
-const IMPORTER_UI_REPO = 'https://github.com/adobe/helix-importer-ui';
 export default class ImportCommand extends AbstractCommand {
   constructor(logger) {
     super(logger);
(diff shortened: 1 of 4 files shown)


trieloff (dismissed):

tripodsan:
> The importer-ui project build process is really slow

you could also pre-build the ui during the _release_ in the ui in helix-importer-ui, i.e. have a semantic-release step that runs the `npm build` and then also commits the `dist` files. then you don't need to run it here.

src/import.cmd.js
```diff
@@ -65,6 +80,35 @@ export default class ImportCommand extends AbstractCommand {
...
+      // clone the ui project
+      await git.clone({
+        fs: fse,
+        http,
+        dir: uiFolder,
+        url: IMPORTER_UI_REPO,
+        ref: 'main',
+        depth: 1,
+        singleBranch: true,
+      });
```
tripodsan, line 101: is this fast? otherwise use just use git, like we did here:
https://github.com/adobe/helix-cli/blob/192f8410af0aff35eb2938080f77fa01cb0a0bb3/src/helix-pages.js#L88-L95

kptdobe: Yes, it is fast. But this I already use `shelljs`, I can remove the `isomorphic-git` dependency. Would do the same anyway.

kptdobe:
> You could also pre-build the ui during the release in the ui in helix-importer-ui, i.e. have a semantic-release step that runs the npm build and then also commits the dist files. then you don't need to run it here.

I like your approach better. Like this, nothing to do after the `git clone` but I can also `git fetch` to get the latest version (would be faster then fetching and potentially have to rebuild everything).

kptdobe:
Thanks to @tripodsan I have added the build process to the semantic release of helix-importer-ui. Now, it is not necessary to build anymore and it is super fast to clone the repo locally or just fetch the latest when `hlx import` starts.

tripodsan (approved):

PR #2078, opened 2022-10-20
fix: add test that shows the unhandled promise rejection

dominique-pfister:
Work in progress: it currently shows the unhandled rejection when building...

### DIFF
--- a/test/import-cmd.test.js
+++ b/test/import-cmd.test.js
@@ -19,7 +19,7 @@ import { fetch } from '../src/fetch-utils.js';
 import ImportCommand from '../src/import.cmd.js';
 
 const TEST_DIR = path.resolve(__rootdir, 'test', 'fixtures', 'import');
-const SAMPLE_HOST = 'https://www.sample.com';
+const SAMPLE_HOST = 'http://www.sample.com';
 
 describe('Integration test for import command', function suite() {
   let nock;
--- a/test/server-livereload.test.js
+++ b/test/server-livereload.test.js
@@ -326,4 +326,32 @@ describe('Helix Server with Livereload', () => {
       await project.stop();
     }
   });
+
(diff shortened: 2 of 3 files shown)


src/server/Indexer.js
```diff
@@ -103,6 +103,10 @@ export default class Indexer {
       pathname,
       response,
     } = this._last;
+    if (!pathname) {
+      log.debug('No last path recorded, dump skipped.');
+      return;
+    }
```
dominique-pfister, line 109: this is the fix 😄

test/import-cmd.test.js
```diff
@@ -19,7 +19,7 @@ import { fetch } from '../src/fetch-utils.js';
 import ImportCommand from '../src/import.cmd.js';
 
 const TEST_DIR = path.resolve(__rootdir, 'test', 'fixtures', 'import');
-const SAMPLE_HOST = 'https://www.sample.com';
+const SAMPLE_HOST = 'http://www.sample.com';
```
dominique-pfister, line 22: switched to `http` because that site will cause a ECONNRESET in the ALPN negotiation phase, which nock doesn't override.

test/server-livereload.test.js
```diff
@@ -326,4 +326,32 @@ describe('Helix Server with Livereload', () => {
       await project.stop();
     }
   });
+
+  it('livereload of a query definition should not cause an unhandled rejection when no last URL exists', async () => {
+    const rejects = [];
+    process.on('unhandledRejection', (reason) => {
+      rejects.push(reason);
+    });
```
dominique-pfister, line 334: interestingly, `mocha` does not display this handled rejection and will not cause the test to fail, either.

tripodsan (approved):

kptdobe (approved):

PR #2088, opened 2022-11-04
Added compression, tested locally 

### LINKED ISSUE #2087: hlx up has no compression
Responses from the helix up server are not compressed which leads to incorrect recommendations from the Google dev tools when developing locally.

For the helix up server to respond with compressed responses if a client indicates that a compressed transfer encoding is supported.

I can live with my own fork if this is not of interest to the community.

**Additional context**
express supports compression middleware, typically enabled with `app.use(compression())` see http://expressjs.com/en/resources/middleware/compression.html for more examples.

Happy to test and PR if of interest. Looks like 1 or 2 lines.

ieb:
Adds compression to helix up removing Chrome dev tools recommendations to compress when developing locally. Fixes #2087.

On by default so no switches required.

Unfortunately the package-lock file was created with an old version of npm, so that got patched also.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "camelcase": "7.0.0",
     "chalk-template": "0.4.0",
     "chokidar": "3.5.3",
+    "compression": "^1.7.4",
     "cookie": "0.5.0",
     "cookie-parser": "1.4.6",
     "dotenv": "16.0.3",
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "camelcase": "7.0.0",
     "chalk-template": "0.4.0",
     "chokidar": "3.5.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -43,6 +43,7 @@
     "camelcase": "7.0.0",
     "chalk-template": "0.4.0",
     "chokidar": "3.5.3",
+    "compression": "^1.7.4",
```
tripodsan, line 46: we usually lock the version, so that there is no surprise when installing the tool.
(although, nowadays, the packages follow semver more strictly than 4-5 years ago).

```suggestion
    "compression": "1.7.4",
```

trieloff: Is that a net-new dependency or just something we turned from transitive to direct?

PR #2091, opened 2022-11-10
fix: do not forward the referer

### LINKED ISSUE #2090: Importer fails with 403 on certain URLs
**Description**
The Franklin importer fails to import https://www.netcentric.biz/sitemap.xml.

**To Reproduce**
1. Launch the Franklin Importer UU
2. Try to crawls https://www.netcentric.biz/sitemap.xml to get the page list

**Expected behavior**
sitemap.xml can be loaded and parse

** Error **
```
warn: Proxy GET request to https://www.netcentric.biz/sitemap.xml: 403 (text/plain)
```
**Version:**
14.14.4

kptdobe:
Fix #2090 

The importer proxy forward the referer, which is always a localhost url. Just removing it unblocks the request from #2090. I even wonder why more of customer imported sites did not complain.

PR is also fixing the failing linting introduced by 645fdfa81c4c7862e9045394fbe0191a7d4df485.

### DIFF
--- a/src/server/HelixServer.js
+++ b/src/server/HelixServer.js
@@ -13,12 +13,11 @@ import { promisify } from 'util';
 import EventEmitter from 'events';
 import path from 'path';
 import express from 'express';
+import compression from 'compression';
 import { fetch } from '../fetch-utils.js';
 import utils from './utils.js';
 import packageJson from '../package.cjs';
 import RequestContext from './RequestContext.js';
-import compression  from 'compression';
-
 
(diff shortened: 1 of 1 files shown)


tripodsan (approved):
maybe it would even be better to rewrite the referrer header to the target host ?

kptdobe:
> maybe it would even be better to rewrite the referrer header to the target host ? 

No clue what's best. Not sure if they typically write some logic on the edge to validate the referrers.

Issue #1865, opened 2021-09-29
Add TLS Options for local development


trieloff:
[`mkcert`](https://github.com/FiloSottile/mkcert) makes it trivially easy to create a root CA in your browser and generate a certificate and key trusted by the new root CA.

We could provide options `--tls-cert localhost.pem` and `--tls-key localhost.key` to bring up the local web server with HTTPS options.

PR #2128, opened 2023-02-02
feat(tls): Support for tls...enables ability to run a local https serv…

bbythewa:
…er...https://github.com/adobe/helix-cli/issues/1865

- [X ] make sure to link the related issues in this description
- [X ] when merging / squashing, make sure the fixed issue references are visible in the commits, for easy compilation of release notes

## Related Issues
https://github.com/adobe/helix-cli/issues/1865

Thanks for contributing!


src/server/utils.js
```diff
@@ -96,6 +96,14 @@ window.LiveReloadOptions = {
...
   https: true,
 };
+</script>`;
+      } else {
+        newbody += `<script>
+window.LiveReloadOptions = { 
+  host: location.hostname, 
+  port: Number(${port}),
+  https: location.protocol,
+};
```
tripodsan, line 106: why this if?

src/server/utils.js
```diff
@@ -263,7 +271,7 @@ window.LiveReloadOptions = {
     ctx.log[level](`Proxy ${req.method} request to ${url}: ${ret.status} (${contentType})`);
 
     const isHTML = ret.status === 200 && contentType.indexOf('text/html') === 0;
-    const injectLR = isHTML && opts.injectLiveReload;
+    const livereload = isHTML && opts.injectLiveReload;
```
tripodsan, line 274: the `port` should be part of the hostname

src/up.cmd.js
```diff
@@ -126,6 +137,28 @@ export default class UpCommand extends AbstractCommand {
       this._project.withCacheDirectory(this._cache);
     }
 
+    if (this._tls) {
+      if (
+        !this._tlsCertPath
+        || !this._tlsKeyPath
+      ) {
+        throw Error(chalk`{redIf using TLS, you must provide both tls cert and tls key...one or both not found`);
```
tripodsan, line 145: typo `redIf`...

src/up.cmd.js
```diff
@@ -126,6 +137,28 @@ export default class UpCommand extends AbstractCommand {
...
+      }
+
+      // this.log.info(chalk`{yellow TLS Enabled...Automatically Switching port to 443`);
+      // this._httpPort = 443;
+
+      // read each file
+      try {
+        const key = fs.readFileSync(this._tlsKeyPath);
+        const cert = fs.readFileSync(this._tlsCertPath);
```
tripodsan, line 154: use `await`

src/up.cmd.js
```diff
@@ -208,7 +241,7 @@ export default class UpCommand extends AbstractCommand {
     this.emit('started', this);
     if (this._open) {
       const url = this._open.startsWith('/')
-        ? `http://localhost:${this._project.server.port}${this._open}`
+        ? `${this._tls ? 'https' : 'http'}://localhost:${this._project.server.port}${this._open}`
```
tripodsan, line 244: I would prefer to set the `scheme` correctly somewhere instead to infer it from the `_tls` everywhere,

tripodsan (changes requested):
Thanks for the PR. It should be restructured a bit and add tests for the new feature.

mokimo:
Small note - the commit message is not formatted correctly it would need to be `feat(tls): ...` to trigger a release your missing the colon `:`

tripodsan:
> Small note - the commit message is not formatted correctly it would need to be `feat(tls): ...` to trigger a release. Your missing the colon `:`

we'll squash and merge and review the commit message anyways...but thank you for spotting it.

src/server/utils.js
bbythewa: I'm not sure what you mean.  

the `if (process.env.CODESPACES === 'true') {` condition was already there previously.  I only added the `else` condition for all other cases since livereload would not work without it.

src/server/utils.js
bbythewa: I'm not sure what you are suggesting. :(

src/up.cmd.js
bbythewa: Sorry, I got the formatting wrong.  I just pushed a fix to this.

src/server/HelixServer.js
```diff
@@ -148,17 +163,35 @@ export default class HelixServer extends EventEmitter {
...
+            reject(new Error(`Error while starting http server: ${err}`));
+          }
+          this._port = this._server.address().port;
+          log.info(`Local Franklin Dev server up and running: http://localhost:${this._port}/`);
+          if (this._project.proxyUrl) {
+            log.info(`Enabled reverse proxy to ${this._project.proxyUrl}`);
+          }
+          resolve();
+        });
+      }
```
tripodsan, line 194: that's a lot of duplicate code. I'm sure we can re-use alot

src/server/utils.js
```diff
@@ -96,6 +96,14 @@ window.LiveReloadOptions = {
   port: 443,
   https: true,
 };
+</script>`;
+      } else {
+        newbody += `<script>
+window.LiveReloadOptions = { 
+  host: location.hostname, 
+  port: Number(${port}),
```
tripodsan, line 104:
```suggestion
  port: ${Number(port)},
```

src/server/utils.js
tripodsan: sorry. must have looked at an outdated change. don't know why I wrote that :-)

src/up.cmd.js
bbythewa: I agree.  This isn't ideal.  At the same time, I wanted to mostly just support the tls feature without having to make people understand they have to set a myriad of other settings (like forgetting to set the "scheme" feature)...

src/server/HelixServer.js
bbythewa: Sure, I'm happy to reduce the code.  Let me push a fix.

src/up.cmd.js
bbythewa: Okay, I see your point and I changed my code to refer to `this._scheme` which will be modified in the case of using tls.

src/server/HelixServer.js
bbythewa: Reformatted code to re-use the listener callback.

bbythewa:
@tripodsan, I think I have fixed all of your suggestions.  Can you re-review? :)

tripodsan (dismissed):

tripodsan:
thanks @bbythewa. looks great. if you could fix the failing tests, that wold be great. after that we can merge it.

bbythewa:
@tripodsan, Thanks for pointing that out...I believe I fixed the failing tests. :)

tripodsan (approved):

Issue #2129, opened 2023-02-04
Add watcher and pre-processing support


sdmcraft:
Frontend code for Franklin Sites using tools such as `less`, `sass/scss`, `babel`, `webpack` etc.  often need an additional pre-processing step to compile these files. [Here's an example](https://github.com/hlxsites/nedbank#run-less-auto-compile-and-franklin-pages-together). Developers across different Franklin Sites projects implement it redundantly and differently.  

This can be improved by adding a generic support in  `helix-cli` for providing an ability to specify watchers and pre-processors purely as project configurations, thus removing the burden of setting up custom implementations for this at project level.

## Motivation
1. With this, the developer need not bother about running/documenting custom scripts which wrap hlx up and/or separate watcher processes. They can simply configure it in package.json once and use the hlx up as usual.

2. This gives more control over the filesystem monitoring aspect of pre-processors by tying them up to the liveReload watcher of the cli, thereby ensuring that all pre-compilation completes before the liveReload happens. We could even ensure the order in which they get executed through configuration. Running these as separate processes may interleave the liveReload and preCompilation processes non-deterministically and introduce race conditions.

Not to be seen as a replacement of grunt or similar setups as that would still be needed. Its primarily their filesystem monitoring aspect that I propose to link to the live-reload of the cli, otherwise keeping it generic. Let me know your thoughts.

## Proposed sample configuration
```
  "watches": [
    {
      "name": "watch:less",
      "globs": [
        "styles/*.less",
        "blocks/**/*.less"
      ],
      "command": "less-watch-compiler --run-once styles styles & less-watch-compiler --run-once blocks blocks"
    }
  ],
```

tripodsan:
I don't think we should have an opinion on how to integrate with developer workflows. Projects with generated code usually already have their grunt or similar setups that can monitor for filesystem changes.

sdmcraft:
> I don't think we should have an opinion on how to integrate with developer workflows. Projects with generated code usually already have their grunt or similar setups that can monitor for filesystem changes.

@tripodsan , indeed that was the approach I adopted in one of the projects which needed `less` pre-compilation. From that experience, a couple of reasons I feel this may be useful:
 
1. With this, the developer need not bother about running/documenting custom scripts which wrap `hlx up`  and/or separate watcher processes. They can simply configure it in `package.json` once and use the `hlx up` as usual.
[...]

PR #2130, opened 2023-02-04
feat: Add watcher and pre-processing support

sdmcraft:
see #2129

## Related Issues

Thanks for contributing!

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "ini": "3.0.1",
     "isomorphic-git": "1.21.0",
     "livereload-js": "3.4.1",
+    "minimatch": "^6.1.6",
     "open": "8.4.0",
     "progress": "2.0.3",
     "rehype-parse": "8.0.4",
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "ini": "3.0.1",
     "isomorphic-git": "1.21.0",
     "livereload-js": "3.4.1",
-    "minimatch": "^6.1.6",
+    "minimatch": "6.1.6",
     "open": "8.4.0",
     "progress": "2.0.3",
     "rehype-parse": "8.0.4",
--- a/src/server/HelixProject.js
+++ b/src/server/HelixProject.js
@@ -172,23 +172,26 @@ export default class HelixProject {
   }
 
   async initWatches() {
-    const pkg = fs.readFileSync(path.resolve(this.directory, 'package.json'));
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -57,6 +57,7 @@
     "ini": "3.0.1",
     "isomorphic-git": "1.21.0",
     "livereload-js": "3.4.1",
+    "minimatch": "^6.1.6",
```
trieloff, line 60: Is this a net new dependency or just a transitive dependency that we are turning into first level?

We also pin all deps by default:

```suggestion
    "minimatch": "6.1.6",
```

src/server/HelixProject.js
```diff
@@ -169,6 +171,30 @@ export default class HelixProject {
...
+    const pkgFile = path.resolve(this.directory, 'package.json');
+    if (fs.existsSync(pkgFile)) {
+      const pkg = fs.readFileSync(pkgFile);
+      const pkgObj = JSON.parse(pkg);
+      const { watches } = pkgObj;
+      if (watches !== undefined) {
+        this.log.debug(`[Glob-Watcher]Registering watches: ${JSON.stringify(watches, null, 2)}`);
+        const watchMap = new Map();
+        watches.forEach((watch) => {
+          if (watch.globs !== undefined && watch.command !== undefined) {
```
trieloff, line 184: This can probably be simplified by not having globs.

trieloff (commented):
Good start. This will become much more clear with a couple of integration tests.

package.json
sdmcraft: Its a new dependency I added for glob matching.

src/server/HelixProject.js
sdmcraft: Looking at [chokidar documentation](https://github.com/paulmillr/chokidar), I couldn't find a way to specify file extensions other than through globs. And I think we need to filter by file extension to avoid unnecessary command executions.

trieloff: Hm. Then 

```json
{
  "scripts": {
    "watch:*.ts": "tscompile",
    "watch:*.less": "lessc"
  }
}
```

could work. One watcher, and the dispatching works based on the glob.

And I'd implement it lazily: no glob support first, and when we hear that it's being used (and slow) we add glob support.

sdmcraft:
Closing the PR for now based on discussion at https://github.com/adobe/helix-cli/issues/2129

PR #2207, opened 2023-06-29
Default to main when `hlx up` is used on detached head state

### LINKED ISSUE #2204: "hlx up" fails if git repo is in detached HEAD state
**Description**
When the repo is in a "detached head" state and the current commit has no named ref, such as when checkout an individual commit or a remote tracking branch, `hlx up` fails with a TypeError: `TypeError: Cannot read properties of undefined (reading 'replace')`.

**To Reproduce**
1. Check out a particular commit either by ref or SHA, e.g. `git checkout HEAD~1`
2. Run `hlx up`
3. See error

**Expected behavior**
The `hlx up` command still runs, and either falls back to the content at the current commit in the remote, _or_ falls back to main, which is the behavior for other cases when the remote ref is not found.

**Screenshots**

image

You can debug the `hlx` command and set a breakpoint in `up.cmd.js` to pause execution on the error. Here's what that looks like for me:

image

**Version:**
1.0.0

**Additional context**
The hlx command _can_ recover when the detached HEAD is on a tag, because a tag is a named ref. In that instance it warns that the remote ref cannot be found, and it falls back to the main branch. That would be the desired behavior for the other detached HEAD states as well.

This seems to be due to an assumption that `GitUtils.getBranch(dir)` will always return a string; however, [the underlying `isomorphic-git` library will return `undefined` if the repo is in a "detached HEAD" state.](https://isomorphic-git.org/docs/en/currentBranch#:~:text=undefined%20if%20the%20HEAD%20is%20detached.)

alexcarol:
Fix #2204

### DIFF
--- a/src/git-utils.js
+++ b/src/git-utils.js
@@ -135,12 +135,13 @@ export default class GitUtils {
 
   /**
    * Returns the name of the current branch. If `HEAD` is at a tag, the name of the tag
-   * will be returned instead, if head is at a commit, main will be returned.
+   * will be returned instead, if head is at a commit, fallback will be returned.
    *
    * @param {string} dir working tree directory path of the git repo
+   * @param {string} fallback fallback value if no branch or tag is found
    * @returns {Promise<string>} current branch or tag
    */
-  static async getBranch(dir) {
+  static async getBranch(dir, fallback = 'main') {
     // current commit sha
     const rev = await git.resolveRef({ fs, dir, ref: 'HEAD' });
     // reverse-lookup tag from commit sha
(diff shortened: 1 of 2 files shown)


tripodsan (changes requested):
I don't think this is the right fix in git-utils. the fact that `main` is the default branch is a convention. hence the fallback should be handled in the helix part.

I would be ok with having a

`getBranch(dir, default = 'main);` 

signature, though :-)

alexcarol:
> I don't think this is the right fix in git-utils. the fact that `main` is the default branch is a convention. hence the fallback should be handled in the helix part.
> 
> I would be ok with having a
> 
> `getBranch(dir, default = 'main);`
> 
> signature, though :-)

Fixed, thanks for the review

tripodsan (approved):

tripodsan:
thanks @alexcarol !

PR #2251, opened 2023-10-02
fix: show deprecation warning and exit

tripodsan:
shows a deprecation warning and exiits:

image

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,7 +13,7 @@
  */
 
 import CLI from './src/cli.js';
-import { checkNodeVersion, validateDotEnv } from './src/config/config-utils.js';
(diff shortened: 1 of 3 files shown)


trieloff:
Not sure if this is the right thing to do. Could we try to depend on aem-cli, so that installing one will pull the other? I'm not sure this will work with -g but it could be worth a try.

trieloff (approved):
@tripodsan I'd count this as a breaking change. So folks who have pinned the major version to their projects will be able to use it (just won't get new functionality), but anyone going through the "normal" `npm install -g` process will get the non-functional version.

So approved, assuming you push a `BREAKING CHANGE` commit to this branch, even if empty

PR #2260, opened 2023-10-18
fix: ensure proper env prefix

### LINKED ISSUE #2259: only use env variables as arguments that are supported
if a user has an env variable for a different application, eg `AEM_COMPANIONPATH`, `aem up` produces an error:

```
unknown argument: companionpath
```

tripodsan:
fixes #2259

### DIFF
--- a/test/fixtures/all.env
+++ b/test/fixtures/all.env
@@ -1,6 +1,9 @@
 #
 # env file for testing that contains all possible and impossible values
 #
+# ignored
+AEMCOMPANIONPATH=foobar
+AEM_COMPANIONPATH=foobar
 
 # global
 AEM_LOG_LEVEL=debug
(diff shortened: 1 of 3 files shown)


test/fixtures/all.env
```diff
@@ -1,6 +1,9 @@
 #
 # env file for testing that contains all possible and impossible values
 #
+# ignored
+AEMCOMPANIONPATH=foobar
+AEM_COMPANIONPATH=foobar
```
dominique-pfister, line 6: silly question: why is `AEM_COMPANIONPATH` ignored?

dominique-pfister (approved):

test/fixtures/all.env
tripodsan: we don't use the logic of yargs to require arguments, but we check it _manually_:
https://github.com/adobe/helix-cli/blob/e274252a6ec8ffadd8460257d383a8504bcbe793/src/cli.js#L36-L44

this was exactly due to this problem of different programs using the same prefix. but I think yargs changed something in this respect, so that `env('AEM')` would match `AEM*` and `AEM_*`. before it was only with the `_`

PR #2352, opened 2024-05-02
fix: support allow-unauthorized in aem up

### LINKED ISSUE #2351: Could add a Options named '--allowUnauthorized ' in command?
I used net environment   in company and i found a problem like this:
image

From the log-level silly , i couldn't to solve this issue with the simple log. XD (Obviously, its about http auth)
And i try to find the reason with the **hlx-cli** source code.
Finally , i found this function named _getFetch()_
image

In my environment   , i need to make the 'allowUnauthorized' parameters  not null.
But I didn't find other way to achieve it besides modify the source code and run it in local .

So i think shall we add a option to control this param easily?

dominique-pfister:
fix #2351

### DIFF
--- a/src/server/BaseServer.js
+++ b/src/server/BaseServer.js
@@ -116,7 +116,7 @@ export class BaseServer extends EventEmitter {
       let retries = 1;
       if (this._project.kill && await utils.checkPortInUse(this._port, this._addr)) {
         try {
-          const res = await getFetch(this._project.allowUnauthorized)(`${this._scheme}://${this._addr}:${this._port}/.kill`);
+          const res = await getFetch(this._project.allowInsecure)(`${this._scheme}://${this._addr}:${this._port}/.kill`);
           await res.text();
         } catch (e) {
           // ignore errors, in case the other server closes connection
--- a/src/server/BaseServer.js
+++ b/src/server/BaseServer.js
@@ -116,7 +116,7 @@ export class BaseServer extends EventEmitter {
       let retries = 1;
       if (this._project.kill && await utils.checkPortInUse(this._port, this._addr)) {
         try {
-          const res = await getFetch(this._project.allowInsecure)(`${this._scheme}://${this._addr}:${this._port}/.kill`);
+          const res = await getFetch(true)(`${this._scheme}://${this._addr}:${this._port}/.kill`);
           await res.text();
         } catch (e) {
           // ignore errors, in case the other server closes connection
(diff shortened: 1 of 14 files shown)


dominique-pfister:
There are more invocations of `getFetch()` in the code, hope this is the only one to fix for this issue.

tripodsan (changes requested):
that's not enough.... search for all `getFetch()` calls :-)

src/up.js
```diff
@@ -92,6 +92,12 @@ export default function up() {
           type: 'string',
         })
         .group(['url', 'livereload', 'no-livereload', 'open', 'no-open', 'print-index', 'cache'], 'AEM Options')
+        .option('allow-unauthorized', {
+          alias: 'allowUnauthorized',
+          describe: 'Whether to allow unauthorized access to server',
```
tripodsan, line 97: maybe this is not a very good description... I think `allowInsecure` would be better

src/server/BaseServer.js
```diff
@@ -116,7 +116,7 @@ export class BaseServer extends EventEmitter {
       let retries = 1;
       if (this._project.kill && await utils.checkPortInUse(this._port, this._addr)) {
         try {
-          const res = await getFetch()(`${this._scheme}://${this._addr}:${this._port}/.kill`);
+          const res = await getFetch(this._project.allowInsecure)(`${this._scheme}://${this._addr}:${this._port}/.kill`);
```
tripodsan, line 119: this is a request back to localhost, so if https, it most probably is a self-signed, to I would use `true` by default

tripodsan (changes requested):
we should probably also add it here:
https://github.com/adobe/helix-cli/blob/ef79a85031259d242919baafd16b28a19a06aa46/src/server/HelixImportServer.js#L104C23-L104C29

IIRC, this was added because some imported sites had invalid certs, but I don't think this should be the default.

@kptdobe WDYT?

kptdobe:
`rejectUnauthorized: false` was introduced by https://github.com/adobe/helix-cli/pull/2040 in the importer server. 
To use the same logic than for `aem up`, `true` can be the default and the option would allow to disable it.

src/up.js
```diff
@@ -92,6 +92,12 @@ export default function up() {
           type: 'string',
         })
         .group(['url', 'livereload', 'no-livereload', 'open', 'no-open', 'print-index', 'cache'], 'AEM Options')
+        .option('allow-insecure', {
+          alias: 'allowInsecure',
+          describe: 'Whether to allow insecure requests to the server',
+          type: 'boolean',
+          default: false,
+        })
```
tripodsan, line 100: maybe also add this to `import.js` ?

tripodsan (approved):

Issue #2426, opened 2024-09-23
aem 'import' should allow a branch


atopper:
**Description**
Switch to a development or new feature branch is awkward when running `aem import`.  The tests imply that it can be done through the `ui-repo` parameter (i.e. **_https://github.com/adobe/helix-importer-ui#somebranch_**) but that does not work.

This would be particularly useful when trying to socialize a new feature or bug fix, and reduce the effort to try it.

**To Reproduce**
1. Be able to run `aem import` successfully
2. Pick a branch in https://github.com/adobe/helix-importer-ui/branches, say 'test'
3. Try to start the importer with that branch: `aem import --ui-repo https://github.com/adobe/helix-importer-ui#test`)
4. Start up fails

**Expected behavior**
The Helix importer UI should start up and change branches to the specified test

**Screenshots**
N/A

**Version:**
Around 16.7.5

**Additional context**
The use of the fragment in the UI isn't publicly advertised, so could be considered a feature.

kptdobe:
You can do that ("manually"):
1. run `aem import`
2. go to `./tools/importer/helix-importer-ui`
3. `git checkout <branch>`
4. build the current branch: `npm run build`
5. reload the importer UI

atopper:
Yes @kptdobe, that's true.  And then the user can add the `--skip-ui` to prevent checking out main again.

I find it is challenging to get people to try a new branch, and if we made it easy and reduce the steps by using _ui-repo_, perhaps even sending them the exact command line to run (i.e. `aem import --ui-repo=https://github.com/adobe/helix-importer-ui#test`), it would be easier for them to quickly try it, and to continue to use.
So it would mainly be for convenience, but I think a good convenience.

arumsey:
I agree with @atopper that the "manual" approach adds too much friction to provide our customers with early access to new features. The aem cli should make it incredibly easy for users of the AEM Importer to access feature branches. The downside would be that it would be too easy to access branches that should not be used and it could encourage the proliferation of even more AEM Importer variations instead of simply focusing on integrating features into the core product.

kptdobe:
I did not realise we already had the `ui-repo` property. Not supporting branches make it useless.

kptdobe:
Fixed with #2432

PR #2432, opened 2024-10-01
#2426 Allow import branch specification

atopper:
Extend the `--ui-repo` parameter to allow a branch specifier, which will make socializing new feature branches easier:
`aem import -ui-repo=https://github.com/adobe/helix-importer-ui#<by-branch>`

## Related Issues
#2426 

@kptdobe I realize you didn't really support this idea, but I went ahead to see the effort involved.

cc @arumsey @catalan-adobe @karlpauls 

Missing from this is an indicator on the UI of the branch, other than 'main', that is being used.  I tried a few ways but thought I would submit this PR without it.  Output from `aem import` is explicit about the branch.

### DIFF
--- a/src/import.cmd.js
+++ b/src/import.cmd.js
@@ -37,18 +37,15 @@ export default class ImportCommand extends AbstractServerCommand {
   }
 
   withUIRepo(value) {
-    this._uiRepo = value;
     this._uiBranch = GitUtils.DEFAULT_BRANCH;
     try {
       const url = new URL(value);
       if (url.hash) {
-        // eslint-disable-next-line prefer-destructuring
-        this._uiRepo = `${url.origin}${url.pathname}`;
-        if (url.hash.length > 1) {
-          // eslint-disable-next-line prefer-destructuring
-          this._uiBranch = url.hash.substring(1);
-        }
+        this._uiBranch = url.hash.substring(1);
       }
+      // Ensure a dangling hash does not create an invalid uiRepo.
+      url.hash = '';
+      this._uiRepo = url.href;
     } catch (e) {
       this.log.error(`Could not process UI Repo correctly: ${value} - ${e.message}`);
       throw e;
@@ -68,27 +65,18 @@ export default class ImportCommand extends AbstractServerCommand {
    * @returns {Promise<void>}
    */
   async handleUIBranch(uiFolder) {
-    try {
-      const branch = await GitUtils.getBranch(uiFolder);
+    const branch = await GitUtils.getBranch(uiFolder);
 
-      if (branch !== this._uiBranch) {
-        this.log.info(
-          chalk`AEM Importer UI was on branch {yellow ${branch}}. Switching to {green ${this._uiBranch}}.`,
-        );
-        // This checkout will produce an error if the uiBranch does not exist (or fails to switch),
-        // will announce the error to the user, and exit the execution.
-        // Error examples:
-        // - "Your local changes to the following files would be overwritten by checkout: <file>"
-        // - "Could not find bad-branch."
-        await git.checkout({
-          fs: fse,
-          http,
-          dir: uiFolder,
-          ref: this._uiBranch,
-        });
-      }
-    } catch (e) {
-      // ignore - fall back to default (main)
+    if (branch !== this._uiBranch) {
+      this.log.info(
+        chalk`AEM Importer UI was on branch {yellow ${branch}}. Switching to {green ${this._uiBranch}}.`,
+      );
+      await git.checkout({
+        fs: fse,
+        http,
+        dir: uiFolder,
+        ref: this._uiBranch,
+      });
     }
   }
 
@@ -111,7 +99,6 @@ export default class ImportCommand extends AbstractServerCommand {
           url: this._uiRepo,
           ref: this._uiBranch,
           depth: 1,
-          singleBranch: true,
         });
 
         this.log.info(`AEM Importer UI is ready. v${await getUIVersion()} (branch: ${this._uiBranch})`);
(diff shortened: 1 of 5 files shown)


src/import.cmd.js
```diff
@@ -37,6 +38,15 @@ export default class ImportCommand extends AbstractServerCommand {
...
     this._uiRepo = value;
+    this._uiBranch = GitUtils.DEFAULT_BRANCH;
+    if (value.includes('#')) {
+      // eslint-disable-next-line prefer-destructuring
+      this._uiRepo = value.split('#')[0];
+      if (value.split('#')[1].length > 0) {
+        // eslint-disable-next-line prefer-destructuring
+        this._uiBranch = value.split('#')[1];
+      }
+    }
```
tripodsan, line 49: use the `URL` class. this also gives you validation for free.

src/import.cmd.js
```diff
@@ -45,11 +55,41 @@ export default class ImportCommand extends AbstractServerCommand {
...
 
+  /**
+   * Ensure the UI branch the caller specified (or defaulted to) is the one being used.
+   * @param uiFolder
+   * @returns {Promise<void>}
+   */
+  async handleUIBranch(uiFolder) {
+    let branch = GitUtils.DEFAULT_BRANCH;
+    try {
+      branch = await GitUtils.getBranch(uiFolder);
```
tripodsan, line 66:
```suggestion
      const branch = await GitUtils.getBranch(uiFolder);
```

src/import.cmd.js
```diff
@@ -45,11 +55,41 @@ export default class ImportCommand extends AbstractServerCommand {
...
+
+      if (branch !== this._uiBranch) {
+        this.log.info(chalk`AEM Importer UI was on branch {yellow ${branch}}. Switching to {green ${this._uiBranch}}.`);
+        // This checkout will produce an error if the uiBranch does not exist (or fails to switch),
+        // will announce the error to the user, and exit the execution.
+        // Error examples:
+        // - "Your local changes to the following files would be overwritten by checkout: <file>"
+        // - "Could not find bad-branch."
```
tripodsan, line 74: but the `git.checkout()` is caught...

src/import.cmd.js
```diff
@@ -61,10 +101,11 @@ export default class ImportCommand extends AbstractServerCommand {
         http,
         dir: uiFolder,
         url: this._uiRepo,
+        ref: this._uiBranch,
         depth: 1,
         singleBranch: true,
       });
-      this.log.info(`AEM Importer UI is ready. v${await getUIVersion()}`);
+      this.log.info(`AEM Importer UI is ready. v${await getUIVersion()} ${this._uiBranch ? `(Branch: ${this._uiBranch})` : ''}`);
```
tripodsan, line 108: `this._uiBranch` is always truthy.... so no need to check

src/import.cmd.js
```diff
@@ -73,13 +114,15 @@ export default class ImportCommand extends AbstractServerCommand {
...
+        ref: this._uiBranch,
         depth: 1,
         singleBranch: true,
         author: {
           name: 'hlx import',
         },
       });
-      this.log.info(`AEM Importer UI is up-to-date. v${await getUIVersion()}`);
+      this.log.info(`AEM Importer UI is up-to-date. v${await getUIVersion()} ${this._uiBranch
+        ? `(Branch: ${this._uiBranch})` : ''}`);
```
tripodsan, line 125: `this._uiBranch` is always truthy.... so no need to check

tripodsan (changes requested):

src/import.cmd.js
```diff
@@ -37,6 +38,22 @@ export default class ImportCommand extends AbstractServerCommand {
...
+    try {
+      const url = new URL(value);
+      if (url.hash) {
+        // eslint-disable-next-line prefer-destructuring
+        this._uiRepo = `${url.origin}${url.pathname}`;
+        if (url.hash.length > 1) {
+          // eslint-disable-next-line prefer-destructuring
+          this._uiBranch = url.hash.substring(1);
+        }
+      }
```
tripodsan, line 51: looks complicated, why not:

```suggestion
      if (url.hash) {
        this._uiBranch = url.hash.substring(1);
        url.hash = '';
        this._uiRepo = url.href;
      }
```

src/import.cmd.js
tripodsan: the comment above does not make sense.... is it a TODO ?

atopper: Took the whole try/catch away and let the exception flow.  I think that simplified it a lot.

src/import.cmd.js
atopper: In the case where a user provided a dangling # (with no branch), the UI Repo was invalid, so I do lines 47 & 48 every time, just in case.

src/import.cmd.js
```diff
@@ -36,7 +37,20 @@ export default class ImportCommand extends AbstractServerCommand {
...
-    this._uiRepo = value;
+    this._uiBranch = GitUtils.DEFAULT_BRANCH;
+    try {
+      const url = new URL(value);
+      if (url.hash) {
+        this._uiBranch = url.hash.substring(1);
+      }
+      // Ensure a dangling hash does not create an invalid uiRepo.
+      url.hash = '';
+      this._uiRepo = url.href;
```
atopper, line 48: ☝️ Always remove hash and set uiRepo.  This also covers the case where `value` has a dangling hash, making the url invalid for future 'git' commands:
`https://github.com/adobe/helix-importer-ui#`

tripodsan (approved):
LGTM, but let's wait for @kptdobe 's feedback, too

kptdobe (approved):
LGTM. Thanks @atopper.

Issue #2498, opened 2025-02-28
Authentication flow doesn't work with Safari


shsteimer:
**Description**
When running `aem up` with an authenticated site. The Auth flow never successfully completes if your default browser is safari. When clicking send on the prompt page, an error occurs which is only surfaced via the dev console.

**To Reproduce**
1. make safari your default browser (my setup: safari 18.3 on Mac OS 15.3.1)
2. run `aem up` for an authenticated site with no stored site token
3. login with the IdP
4. when you arrive at `https://admin.hlx.page/auth/adobe/ack?code=...` click send
5. note that nothing happens
6. open dev tools console and note the error

**Expected behavior**
It should work ;)

**Screenshots**

errors from console:
```
[Warning] [blocked] The page at https://admin.hlx.page/auth/adobe/ack?code=.... requested insecure content from http://localhost:3000/.aem/cli/login/ack. This content was blocked and must (ack, line 7)

[Error] Not allowed to request resource
	(anonymous function) (ack:7)
	sendPost (ack:6)
	(anonymous function) (ack:23)
[Error] Fetch API cannot load http://localhost:3000/.aem/cli/login/ack due to access control checks.
	(anonymous function) (ack:7)
	sendPost (ack:6)
	(anonymous function) (ack:23)
[Error] Unhandled Promise Rejection: TypeError: Load failed
	(anonymous function) (ack:7)
```

**Version:**
```
ssteimer@Seans-M3-Macbook-Pro ups % aem --version
16.10.1
```

andreituicu:
It looks like Safari is the only browser that very strictly enforces not making http calls from a secure https connection, even on localhost.
The other browsers treat localhost like a secure https connection.

https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content#loading_locally_delivered_mixed-resources
https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content#browser_compatibility

Will probably need to switch the response from Helix Admin to transfer information via a redirect to localhost (assuming that's supported by Safari, will need to check).
But that has different security disadvantages, unless we use something like PKCE, which requires more effort to implement.
Will need to check what we can do...

shsteimer:
@andreituicu While obviously we should strive to support all major browsers, I think it would be sufficient, or at least a good stop gap, to detect the user agent is using safari, and ask them to use a different browser to authenticate

andreituicu:
[...]

PR #2500, opened 2025-03-03
fix: Give a warning when using Safari for CLI login

andreituicu:
## Related Issues
#2498

### DIFF
--- a/src/server/HelixServer.js
+++ b/src/server/HelixServer.js
@@ -47,6 +47,18 @@ export class HelixServer extends BaseServer {
   }
 
   async handleLogin(req, res) {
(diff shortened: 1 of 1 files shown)


shsteimer:
The message itself I think is fine.

I get the hesitation on such warnings, and open to the idea that the a temp fix like this isn't worth the trouble. Relatively small fraction of projects use site auth. Relatively small fraction of users use safari on desktop. It's entirely plausible this fix has an audience of one person (me). Given that, I'd be supportive of holding off and seeing if we get more reports of the problem before we do anything.

tripodsan (approved):

Issue #2520, opened 2025-04-17
Provide version information about the current aem-cli


bhellema:
To provide a better experience for end users, expose a /version endpoint which can be used to determine if the client is running an old version of the aem-cli.

tripodsan:
duplicate of https://github.com/adobe/helix-cli/issues/2225

PR #2521, opened 2025-04-17
fix(2520): Provide version information about the current aem-cli

bhellema:
## Related Issues
fix: #2520

### DIFF
--- a/src/server/BaseServer.js
+++ b/src/server/BaseServer.js
@@ -63,7 +63,7 @@ export class BaseServer extends EventEmitter {
       res.send('Goodbye!');
       this.stop();
     });
-    this._app.get('/version', async (req, res) => {
+    this._app.get('/.hlx/version', async (req, res) => {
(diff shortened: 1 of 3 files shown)


src/server/BaseServer.js
```diff
@@ -63,6 +63,13 @@ export class BaseServer extends EventEmitter {
       res.send('Goodbye!');
       this.stop();
     });
+    this._app.get('/version', async (req, res) => {
```
tripodsan, line 66: I prefer a _namespaced_ route, like `/.hlx/version`

tripodsan (changes requested):
avoid url space poisoning.

src/server/BaseServer.js
bhellema: Out of curiosity, why place the version information under .hlx?  Is not not common practice for services to provide a basic /version endpoint?

tripodsan (approved):

PR #2575, opened 2025-08-28
chore(README): move hlx install conflict to troubleshooting section

alexkli:
Improving the readme: Moving the `File exists: /opt/homebrew/bin/hlx` error description to the troubleshooting section at the bottom.

Reason: 
- running into installation conflict with having an older `@adobe/helix-cli` installed should become less of an issue over time now
- should not occupy so much room at the top of the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,22 +17,7 @@ Install `aem` as a global command. You need Node 12.11 or newer.
 npm install -g @adobe/aem-cli
 ```
 
-### Migrating from older version
-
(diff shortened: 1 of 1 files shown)


alexkli:
BTW I would argue that one should rename this git repository to `adobe/aem-cli` as well, to avoid confusion. Github should still redirect from the old repository name.

tripodsan (approved):

PR #2578, opened 2025-08-29
fix(up): proxy all cookie and set-cookie headers

alexkli:
## Use case

Support passing all cookies while proxying, in both directions, e.g. for custom authentication.

_**Update after I created the PR:** Originally I was trying to run `aem up` aka `localhost:3000` as proxy before a custom cloudflare worker where we do authentication using cookies. To make that authentication work, `aem up` needs to proxy cookies in both directions, which is what this PR enabled._

_But in the meantime I figured it's better to simulate the real setup where the cloudflare worker runs in front of Helix. Hence I am running the cloudflare worker (`wrangler dev`) locally now _in front_ of EDS, using `localhost:3000` as origin for Helix requests, so it can talk to `aem up` running locally as well. See https://github.com/aemsites/koassets/pull/32. For that I don't need this PR._

## Changes
1. New `--cookies` option to proxy all cookies in the requests
   - In the current implementation, only the `hlx-auth-token` cookie on requests is proxied through. I could not find a reason why but I assume there is one.
   - That's why I kept the default behavior and introduced this new option.
   - **QUESTION: If there is no need for the old behavior**, and cookies can always be passed in requests, the change would be a lot simpler and the new `--cookies` option would not be required.
2. Support multiple `Set-Cookie` headers in responses
    - These were previously incorrectly joined into a single `Set-Cookie` header, which browsers see as a single cookie, leading to incorrect results.
    - I don't think that was intentional. It's a technical limitation of the fetch Response API, which requires a special handling in node-fetch: https://www.npmjs.com/package/node-fetch#extract-set-cookie-header

Also added unit tests.

## New option

```
> aem up --help
...

Options:
...

  --cookies                          Proxy all cookies in requests. By default,
                                     only the hlx-auth-token cookie is proxied.
                                                      [boolean] [default: false]
```

### DIFF
--- a/src/server/HelixProject.js
+++ b/src/server/HelixProject.js
@@ -68,6 +68,11 @@ export class HelixProject extends BaseProject {
     return this;
   }
 
+  withCookies(value) {
+    this._server.withCookies(value);
+    return this;
+  }
+
   get proxyUrl() {
     return this._proxyUrl;
   }
--- a/src/server/HelixServer.js
+++ b/src/server/HelixServer.js
@@ -34,6 +34,7 @@ export class HelixServer extends BaseServer {
     this._enableLiveReload = false;
     this._app.use(compression());
     this._autoLogin = true;
+    this._cookies = false;
   }
 
   withLiveReload(value) {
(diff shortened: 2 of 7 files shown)


alexkli:
I might actually not need this anymore... see the **update** section in the description above.

tripodsan:
I think having the additional option is better

alexkli:
@tripodsan With the new `--cookies` option this is probably `feat` rather than a `fix`?

PR #2707, opened 2026-04-15
feat: extend --html-folder to serve local JSON form definitions on-the-fly

navneet-ag:
Extends the existing `--html-folder` mechanism to also resolve `.json` files,
rendering them as AEM form block HTML on-the-fly without requiring a publish step.

When a request hits the html-folder handler and no `.html` or `.plain.html` file
is found, it falls back to a `.json` file and calls `generateFormPageHtml()` to
produce the standard form block structure the client-side form script expects.

Resolution priority: `.html` > `.plain.html` > `.json`

**Usage:**
```
aem up --html-folder forms --html-mount /content/forms/af/
```

Replaces the previously proposed `--local-forms` / `--local-forms-mount` flags
based on reviewer feedback.

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/server/HelixServer.js
+++ b/src/server/HelixServer.js
@@ -232,26 +232,6 @@ export class HelixServer extends BaseServer {
       // Neither exists
     }
 
-    // Try .json (form definition files — rendered as form block HTML on-the-fly)
-    const jsonFile = path.resolve(
-      this._project.directory,
-      this._htmlFolder,
-      `${relativePath}.json`,
-    );
-
-    if (!utils.validatePathSecurity(jsonFile, this._project.directory)) {
-      return null;
-    }
-
-    try {
-      const stats = await lstat(jsonFile);
-      if (stats.isFile()) {
-        return { file: jsonFile, isJson: true };
-      }
-    } catch (e) {
-      // .json not found either
-    }
-
     return null;
   }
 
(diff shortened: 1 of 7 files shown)


tripodsan:
this use case can be covered with the  `--html-folder`. I don't think we should add markup source provider specific rendering in this CLI.

tripodsan (changes requested):
json is used to store sheets and should not be rendered differently. I don't see why the server should transform it into html

navneet-ag:
@tripodsan  Agreed — removed the JSON→HTML transform completely. The server no longer renders JSON differently.    
                                                              
  The only remaining change: --html-folder now serves static files (like .json) as-is from the folder. This enables a local
   form dev workflow where the HTML uses ```<a href="/forms/def.json">```  and the form block JS fetches the JSON
   at runtime. I can edit the JSON and see changes immediately without touching the HTML.

src/server/HelixServer.js
```diff
@@ -325,6 +329,55 @@ export class HelixServer extends BaseServer {
...
 
+  /**
+   * Serves static files from the HTML folder as-is
+   * @param {Express.Request} req request
+   * @param {Express.Response} res response
+   * @param {Function} next next middleware
+   * @param {string} relativePath path relative to HTML folder
+   */
+  async serveHtmlFolderStaticFile(req, res, next, relativePath) {
+    if (relativePath.includes('/../') || relativePath.includes('..')) {
```
tripodsan, line 340: why do you need both checks ? `isn't `..` sufficient?

tripodsan (approved):

tripodsan:
merge with #2717

PR #2709, opened 2026-04-19
feat(content): clone HTML/JSON by default, add --media for assets

davidnuescheler:
## Summary
- `aem content clone` now downloads only **`.html` and `.json`** files by default (case-insensitive).
- Add **`--media`** to include all other file types (images, video, SVG, PDF, etc.), matching the previous all-files behavior.
- When the remote lists more files than are cloned, the CLI logs the discovered vs cloned counts and mentions `--media`.
- Large-clone confirmation uses the **filtered** file count.

## Testing
- Unit tests updated in `test/content/clone.cmd.test.js` and `test/content/content-commands.test.js`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/content/content-shared.js
+++ b/src/content/content-shared.js
@@ -10,6 +10,8 @@
  * governing permissions and limitations under the License.
  */
 
+import path from 'path';
+
 export const CONTENT_DIR = 'content';
 export const CONFIG_FILE = '.da-config.json';
 export const GIT_AUTHOR = { name: 'aem-cli', email: 'aem-cli@adobe.com' };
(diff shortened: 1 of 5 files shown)


kptdobe (dismissed):

src/content/content-shared.js
```diff
@@ -22,6 +24,34 @@ export const LARGE_CLONE_FILE_THRESHOLD = 10000;
...
+ * @param {{ ext?: string, name?: string, path?: string }} file
+ * @returns {string} lower-case extension without leading dot, or '' if unknown
+ */
+function contentFileExtension(file) {
+  if (file.ext != null && String(file.ext) !== '') {
+    return String(file.ext).replace(/^\./, '').toLowerCase();
+  }
+  const base = file.name || file.path || '';
+  return path.extname(base).replace(/^\./, '').toLowerCase();
+}
```
tripodsan, line 37: why so complicated? just include the `.` in the extension.

src/content/content-shared.js
```diff
@@ -22,6 +24,34 @@ export const LARGE_CLONE_FILE_THRESHOLD = 10000;
...
+
+/**
+ * When `includeMedia` is false, keeps only `.html` and `.json` entries (da.live list items).
+ * @param {Array<{ ext?: string, name?: string, path?: string }>} files
+ * @param {boolean} includeMedia
+ * @returns {typeof files}
+ */
+export function filterFilesForContentClone(files, includeMedia) {
```
tripodsan, line 45: why is the check for media here? it just returns `true`, so why filter at all?
but it's ok to keep it, in case in the future there are more files to filter.

tripodsan (changes requested):
improve extension operation (this certainly exists already somewhere....)

PR #2711, opened 2026-04-20
fix: remove if-modified-since from proxy request to ensure that content-type is included

tripodsan:


kptdobe (approved):
Can you explain why / when this is a problem ?

tripodsan:
> Can you explain why / when this is a problem ?

I debugged an issue where modifications to the local `head.html` where not reflected. the issue was that the response from aem.page did not include a `content-type` header and so it bypassed the head.html injection. 

I saw that the response was a `304`, since the browser included a `if-modified-since`. I think on a `304` it is ok for the CDN to omit the `content-type` header. (most of the devs work with cache-disabled (network tab) and then chrome doesn't send the `if-modified-since` , so I think they rarely get this issue)

PR #2724, opened 2026-05-12
fix: Add --preferPlainHtml flag

swetabar:
Add an opt-in --prefer-plain-html flag to aem up so .plain.html wins over .html sibling in --html-folder mode

## Related Issues
https://github.com/Adobe-AEM-Foundation/aem-experience-catalyst/issues/1129

### DIFF
--- a/src/server/HelixServer.js
+++ b/src/server/HelixServer.js
@@ -205,6 +205,35 @@ export class HelixServer extends BaseServer {
   }
 
   /**
+   * Resolves a candidate file inside the HTML folder, returning its absolute path
+   * if it exists and passes the security check, or null otherwise.
+   * @param {string} relativePath path relative to HTML folder (without extension)
+   * @param {string} ext extension to append (e.g. '.html', '.plain.html')
+   * @returns {Promise<string|null>} absolute path, or null if missing/invalid
+   */
+  async resolveCandidate(relativePath, ext) {
+    const file = path.resolve(
+      this._project.directory,
+      this._htmlFolder,
+      `${relativePath}${ext}`,
+    );
+
+    if (!utils.validatePathSecurity(file, this._project.directory)) {
+      return null;
+    }
+
+    try {
+      const stats = await lstat(file);
+      if (stats.isFile()) {
+        return file;
+      }
+    } catch (e) {
+      // not found
+    }
+    return null;
+  }
+
+  /**
    * Resolves which HTML file to serve from the HTML folder
    * @param {string} relativePath path relative to HTML folder
    * @returns {Promise<{file: string, isPlain: boolean}|null>} resolved file info or null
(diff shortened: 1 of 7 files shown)


src/server/HelixServer.js
```diff
@@ -215,41 +220,45 @@ export class HelixServer extends BaseServer {
...
       }
     } catch (e) {
-      // .html not found, try .plain.html
+      // first extension not found, try second
     }
 
-    // Try .plain.html
-    const plainHtmlFile = path.resolve(
+    // Try second extension
+    const secondFile = path.resolve(
```
alexcarol: rather than first/second have you considered an approach where an arbitrary length array of options is allowed? (if it helps simplify, otherwise probably not worth it)

tripodsan:
@davidnuescheler @kptdobe why is there a fallback to `.plain.html` at all?

kptdobe:
Because you approved it - see https://github.com/adobe/helix-cli/pull/2631
@karlpauls introduced it, not super clear why.

alexcarol:
In experience modernization agent we're generating .plain.html that are both used to show the user a "preview" of their content *and* then get uploaded to DA once the user is satisfied. Considering the new "aem content" approach it might make sense to look at going "back" to generating .md files, but for now we're still using ".plain.html", so it makes sense to have a way to have a way to "prioritize them".

tripodsan:
> Because you approved it - see #2631 @karlpauls introduced it, not super clear why.

@karlpauls I'm not sure anymore why we need to serve the `.plain.html` files on non plain requests...

karlpauls:
The goal was to serve local content as it would come from the edge.

Originally, the html folder feature only served full HTML files (including a `<head>`) - just without the .html extension. This required authors and agents to produce complete HTML documents locally, which is cumbersome, and made updating the `<head>` painful since every file would need to be regenerated.

With #2631, we added support for serving a plain.html file, matching what .page/index.plain.html would return from the edge. This is simpler to produce and allows the head to be injected/managed separately.

For backwards compatibility, a .html file takes precedence over a .plain.html file. Additionally, no injection is applied to .html files - injection only happens for .plain.html.

src/server/HelixServer.js
tripodsan: I agree, create a helper function that resolves the local file and iterate over the candiates.

tripodsan (changes requested):

src/server/HelixServer.js
```diff
@@ -205,63 +205,60 @@ export class HelixServer extends BaseServer {
   }
 
   /**
-   * Resolves which HTML file to serve from the HTML folder
-   * @param {string} relativePath path relative to HTML folder
-   * @returns {Promise<{file: string, isPlain: boolean}|null>} resolved file info or null
+   * Resolves a candidate file inside the HTML folder, returning its absolute path
+   * if it exists and passes the security check, or null otherwise.
+   * @param {string} relativePath path relative to HTML folder (without extension)
+   * @param {string} ext extension to append (e.g. '.html', '.plain.html')
```
tripodsan: the relativePath / ext separation is not important to this function, so I would only use 1 argument: `relativePath` and concat them in the caller.

src/server/HelixServer.js
```diff
@@ -205,63 +205,60 @@ export class HelixServer extends BaseServer {
...
+  /**
+   * Resolves which HTML file to serve from the HTML folder
+   * @param {string} relativePath path relative to HTML folder
+   * @returns {Promise<{file: string, isPlain: boolean}|null>} resolved file info or null
+   */
+  async resolveHtmlFolderFile(relativePath) {
+    // Security check: prevent path traversal with /../ anywhere in the path
+    if (relativePath.includes('/../')) {
+      return null;
+    }
```
tripodsan: this is also checked in the `validatePathSecurity` above... so no need to check here,

tripodsan (changes requested):

tripodsan (approved):
thanks. I'll merge it asap

tripodsan:
merged with #2728

PR #2733, opened 2026-05-27
fix: properly serve .plain.html requests for documents in local content directory

### LINKED ISSUE #2732: .plain.html requests not correctly handled for local content files
**Description**
when creating local html documents under the cloned aem content directory, .plain.html requests to these files are not served, resulting in 404s.

**To Reproduce**
1. run `aem content clone --all`
2. create a new file at `content/nav.html` and add nav markup to the file to it
3. run `aem up`
4. make a request to nav.plain.html and note that it 404s

**Expected behavior**
the plain html should be served, as it would be once the content is pushed and previewed

**Version:**
16.19.8

shsteimer:
fix: #2732

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/server/HelixServer.js
+++ b/src/server/HelixServer.js
@@ -467,7 +467,6 @@ export class HelixServer extends BaseServer {
               : contentFilePath;
             let htmlContent = await readFile(servedFilePath, 'utf-8');
             if (isPlainFallback) {
-              // .plain.html callers want the raw fragment — extract <main> content
               if (liveReload) {
                 liveReload.registerFile(ctx.requestId, servedFilePath);
               }
--- a/src/server/HelixServer.js
+++ b/src/server/HelixServer.js
@@ -476,15 +476,16 @@ export class HelixServer extends BaseServer {
               }
             }
             if (isPlainFallback) {
-              // .plain.html callers want the raw body fragment — skip head wrapping
+              // .plain.html callers want the raw fragment — extract <main> content
               if (liveReload) {
                 liveReload.registerFile(ctx.requestId, servedFilePath);
               }
+              const fragment = utils.extractMainContent(htmlContent);
               res.set({
                 'content-type': 'text/html; charset=utf-8',
                 'access-control-allow-origin': '*',
               });
-              res.send(htmlContent);
+              res.send(fragment);
               log.debug(`${pfx}served from ${CONTENT_DIR}/: ${ctx.path}`);
               return;
             }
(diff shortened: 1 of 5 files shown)


src/server/HelixServer.js
```diff
@@ -459,7 +459,27 @@ export class HelixServer extends BaseServer {
         try {
           if (contentFilePath.endsWith('.html')) {
             // readFile throws EISDIR for directories and ENOENT for missing files
-            let htmlContent = await readFile(contentFilePath, 'utf-8');
+            // .plain.html is a virtual URL convention — AEM never stores .plain.html files,
+            // so always read the corresponding .html file and serve the <main> fragment.
+            const isPlainFallback = contentFilePath.endsWith('.plain.html');
```
shsteimer, line 464: worth noting that this assumes there will never be a .plain.html file in the content checkout directory. If that assumption doesn't hold, then this will need to be adapted to handle that case.

src/server/utils.js
```diff
@@ -787,6 +790,21 @@ window.LiveReloadOptions = {
...
     return `<html><head>${fullHead}</head><body><header></header><main>${content}</main><footer></footer></body></html>`;
   },
+
+  /**
+   * Extracts the innerHTML of the <main> element from a full HTML document.
+   * Returns the original content unchanged if no <main> is found.
+   * @param {string} html full HTML document
+   * @returns {string} innerHTML of <main>, or original html if no <main> present
+   */
+  extractMainContent(html) {
```
kptdobe, line 800: Missing unit tests

src/server/HelixServer.js
```diff
@@ -459,7 +459,27 @@ export class HelixServer extends BaseServer {
...
+            if (isPlainFallback) {
+              // .plain.html callers want the raw fragment — extract <main> content
+              if (liveReload) {
+                liveReload.registerFile(ctx.requestId, servedFilePath);
+              }
+              const fragment = utils.extractMainContent(htmlContent);
+              res.set({
+                'content-type': 'text/html; charset=utf-8',
+                'access-control-allow-origin': '*',
+              });
```
kptdobe, line 477: why is `access-control-allow-origin` needed ? this is not happening in prod...

kptdobe (changes requested):

src/server/HelixServer.js
shsteimer: It looks to me like we are setting `access-control-allow-origin: *` on all responses from the local dev server, originally added in #1900. so this just follows that same pattern as all other locally-served responses in this handler (lines 512, 521, 541, 551).

src/server/utils.js
shsteimer: tests added

kptdobe (approved):

PR #2735, opened 2026-06-03
fix(content): normalize org/repo to lowercase in aem content clone

kptdobe:
## Summary

- `aem content clone --all` silently skipped all files when the GitHub remote used mixed-case org/repo names (e.g. `kptdobe/Repo-DA`)
- DA admin always returns paths with lowercase org/repo (`/kptdobe/repo-da/...`), so the case-sensitive `startsWith` prefix check failed for every file
- Fix: lowercase `org` and `repo` immediately after extracting them from the git URL, before any DA API usage or path comparisons

## Test plan

- [ ] New unit test covers the case mismatch scenario: git remote returns `myOrg/MyRepo`, DA paths come back as `/myorg/myrepo/...`, files are downloaded correctly and config stores lowercase values
- [ ] All 23 existing `CloneCommand` tests still pass
- [ ] `npm run lint` passes

### DIFF
--- a/src/content/clone.cmd.js
+++ b/src/content/clone.cmd.js
@@ -101,14 +101,14 @@ export default class CloneCommand {
       throw new Error('Clone root path was not set (internal error).');
     }
 
-    // 1. Resolve org/repo from git remote
+    // 1. Resolve owner/repo from git remote
     const originUrl = await GitUtils.getOriginURL(this._dir);
     if (!originUrl) {
       throw new Error('No git remote found. Run `aem content clone` inside an AEM project directory.');
     }
-    const org = originUrl.owner.toLowerCase();
+    const owner = originUrl.owner.toLowerCase();
     const repo = originUrl.repo.toLowerCase();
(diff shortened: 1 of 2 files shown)


tripodsan (approved):

src/content/clone.cmd.js
```diff
@@ -106,8 +106,8 @@ export default class CloneCommand {
     if (!originUrl) {
       throw new Error('No git remote found. Run `aem content clone` inside an AEM project directory.');
     }
-    const org = originUrl.owner;
-    const { repo } = originUrl;
+    const org = originUrl.owner.toLowerCase();
+    const repo = originUrl.repo.toLowerCase();
```
tripodsan: why is this called `org/repo` and not `owner/repo` or `org/site` ? would be good to be consistent,

tripodsan (dismissed):

src/content/clone.cmd.js
kptdobe: Arff.. we are not done with this mix! Renamed to orgin convention (owner/repo).

PR #2738, opened 2026-06-04
feat(content): add --org/--site options to aem content clone

### LINKED ISSUE #2725: [content] support repoless projects
I would like the aem-cli to support repoless projects.

**Additional context**
It appears that despite manually editing `.da-config.json`, trying to get content when it is not exactly the same as the code `org/repo`.

kptdobe:
## Summary

Closes #2725 — supports repoless projects by allowing the da.live org and site to be specified explicitly on the CLI instead of being derived from the git remote.

```
aem content clone --org <org> --site <site> --all
aem content clone --org adobe --site www --path /en
```

- `--org` and `--site` must be provided together; either one alone is rejected with a clear error
- When both are provided the git remote lookup is skipped entirely, so the command works in directories with no GitHub origin
- Values are normalized to lowercase (consistent with git-derived resolution)
- When neither is provided, existing behavior is unchanged

## Test plan

- [ ] `withOrg` / `withSite` builder methods set internal state
- [ ] `withOrg` / `withSite` return `this` for chaining
- [ ] Clone succeeds with `--org`/`--site` and no git remote
- [ ] Config stores lowercased org/site when mixed-case values are passed
- [ ] Existing behavior (git remote resolution) unchanged when `--org`/`--site` are omitted
- [ ] Error thrown when no git remote and `--org`/`--site` not provided

### DIFF
--- a/src/content/clone.cmd.js
+++ b/src/content/clone.cmd.js
@@ -97,12 +97,12 @@ export default class CloneCommand {
   }
 
   withOrg(org) {
-    this._org = (org || '').trim() || null;
+    this._org = org ? org.trim() : null;
     return this;
   }
 
   withSite(site) {
-    this._site = (site || '').trim() || null;
+    this._site = site ? site.trim() : null;
     return this;
   }
 
--- a/src/content/clone.cmd.js
+++ b/src/content/clone.cmd.js
@@ -114,9 +114,14 @@ export default class CloneCommand {
     }
 
     // 1. Resolve org/site: explicit --org/--site override, otherwise derive from git remote
+    const hasOrg = this._org != null && this._org !== '';
+    const hasSite = this._site != null && this._site !== '';
+    if (hasOrg !== hasSite) {
+      throw new Error('--org and --site must be used together.');
+    }
     let org;
     let site;
-    if (this._org && this._site) {
+    if (hasOrg && hasSite) {
       org = this._org.toLowerCase();
       site = this._site.toLowerCase();
     } else {
--- a/src/content/clone.cmd.js
+++ b/src/content/clone.cmd.js
@@ -97,12 +97,12 @@ export default class CloneCommand {
   }
 
   withOrg(org) {
-    this._org = org;
+    this._org = org != null ? String(org).trim() || null : null;
     return this;
   }
 
   withSite(site) {
-    this._site = site;
+    this._site = site != null ? String(site).trim() || null : null;
     return this;
   }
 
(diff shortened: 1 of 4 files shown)


src/content/clone.cmd.js
```diff
@@ -94,20 +96,42 @@ export default class CloneCommand {
     return this;
   }
 
+  withOrg(org) {
+    this._org = org != null ? String(org).trim() || null : null;
```
tripodsan: why so complicated? I'd assume that yargs trims the arguments.

src/content/clone.cmd.js
```diff
@@ -94,20 +96,42 @@ export default class CloneCommand {
...
+    return this;
+  }
+
+  withSite(site) {
+    this._site = site != null ? String(site).trim() || null : null;
```
tripodsan: same

src/content/clone.cmd.js
```diff
@@ -94,20 +96,42 @@ export default class CloneCommand {
...
     if (this._rootPath == null) {
       throw new Error('Clone root path was not set (internal error).');
     }
 
-    // 1. Resolve org/site from git remote (GitHub owner/repo → da.live org/site)
-    const originUrl = await GitUtils.getOriginURL(this._dir);
-    if (!originUrl) {
-      throw new Error('No git remote found. Run `aem content clone` inside an AEM project directory.');
+    // 1. Resolve org/site: explicit --org/--site override, otherwise derive from git remote
+    const hasOrg = this._org != null && this._org !== '';
```
tripodsan:
```suggestion
    const hasOrg = !!this._org;
```

src/content/clone.cmd.js
```diff
@@ -94,20 +96,42 @@ export default class CloneCommand {
...
+    const hasSite = this._site != null && this._site !== '';
```
tripodsan:
```suggestion
    const hasSite = !!this._site;
```

src/content/clone.js
```diff
@@ -47,6 +55,11 @@ export default function clone() {
           default: false,
         })
         .check((argv) => {
+          const org = argv.org != null ? String(argv.org).trim() : '';
+          const site = argv.site != null ? String(argv.site).trim() : '';
+          if ((org && !site) || (!org && site)) {
+            return '--org and --site must be used together.';
```
tripodsan: why is it checked here and in `clone.cmd.js` ? I think it's enough to validate here.

also better:
```
const org = (argv.org || '').trim();
```

tripodsan (changes requested):
- remove extra check in clone.cmd.js (rely on yargs validation)
- simplify

kptdobe:
Simplified.
Initially, if we wanted to have the .cmd behave the same, we needed more complexity.

tripodsan (approved):

PR #2744, opened 2026-06-13
fix(content): use multipart POST to upload files to DA

### LINKED ISSUE #2743: bug(content): aem content push silently drops JSON file changes
## Summary

Running `aem content push` after editing a JSON file logs `✓` and advances the synced ref, but the file in DA is unchanged.

## Root Cause

`putSource` in `src/content/da-api.js` sends all files via `PUT` with a raw request body and `Content-Type` set to the file's MIME type (e.g. `application/json` for `.json` files).

The DA admin API (`admin.da.live`) only accepts raw body uploads for `text/html`. For `application/json` (and every other content type), `putHelper` in da-admin returns `undefined`. `putObject` then writes an empty `{}` to the file's props sidecar and skips the content file entirely. It still returns `201` with a valid JSON body, so helix-cli treats the push as successful and advances the synced ref. The file change is silently dropped.

This appears to be intentional da-admin behavior. The [test suite asserts it](https://github.com/adobe/da-admin/blob/main/test/helpers/source.test.js#L122):

```js
it('Returns undefined for application/json raw body', async () => {
  // ...
  assert.strictEqual(helped, undefined);
});
```

## What the DA API Expects

The DA API docs (https://docs.da.live/developers/api/source) require multipart/form-data POST for all file types:

```bash
# JSON file
curl -X POST \
  'https://admin.da.live/source/org/site/data.json' \
  --header 'Authorization: Bearer {IMS_TOKEN}' \
  --form 'data=@/path/to/data.json'
```

The server accepts both `PUT` and `POST` (same handler), so the HTTP method is not the issue. The body format is.

## Impact

Any non-HTML file pushed via `aem content push` silently fails to update in DA.

## Fix

Update `putSource` to use `POST` + `multipart/form-data` for all file types, wrapping the buffer in a `Blob` in the `data` field. The `text/html` raw body path is an undocumented server-side shortcut and should not be the basis for how helix-cli uploads files.

shsteimer:
## Related Issues

Fixes #2743.

`DaClient.postSource` (previously `putSource`) now sends files as `multipart/form-data POST` with the content as a `Blob` in the `data` field, matching the [documented DA API contract](https://docs.da.live/developers/api/source). The previous raw-body `PUT` with `Content-Type: application/json` was silently ignored by the server for all non-HTML files.

The rename from `putSource` → `postSource` is internal — `DaClient` is not exported from the package.

Note: potentially related to #2709, which adds JSON to the default clone set.

---

Thanks for contributing!

### DIFF
--- a/src/content/da-api.js
+++ b/src/content/da-api.js
@@ -140,7 +140,10 @@ export class DaClient {
   }
 
   /**
-   * Uploads a file via PUT.
+   * Uploads a file via multipart POST — the documented DA API contract.
+   * Raw body uploads are only supported by DA for text/html; all other types
+   * (including application/json) are silently ignored and must use form-data.
+   * See: https://docs.da.live/developers/api/source
    * @param {string} org
    * @param {string} site
    * @param {string} daPath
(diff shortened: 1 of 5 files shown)


tripodsan (changes requested):
I'd rather change the server to finally allow non multipart puts....

shsteimer:
closing in favor of https://github.com/adobe/da-admin/pull/294

PR #2754, opened 2026-07-03
fix(server): auth to da.live preview host for content.da.live images

### LINKED ISSUE #2752: [content] Missing images after clone and on aem up
- In DA, create a page and drag&drop an image
- `aem content clone` the project
- run `aem up`
- open the page locally
Image is missing (`401`).

The reason: in the page markup, image points to https://content.da.live/ which has been secured last week (for images).

cc @andreituicu  @davidnuescheler

kptdobe:
## Summary
- `content.da.live` image URLs in local `content/` pages are rewritten to the site's `*.preview.da.live` host, since `content.da.live` is auth-gated and not reachable during local dev.
- Since that preview host also requires an IMS-authenticated cookie, a client-side bootstrap is injected (only on pages that reference it): it first checks for an existing valid cookie by probing an actual asset (not the site root, which is often served regardless of auth), and only if that fails loads IMS / redirects through a new `/.aem/cli/da-login` dev-server route (the `darkalley` IMS client only allows its fixed `:9898` callback as `redirect_uri`, so a plain client-side `signIn()` redirects to the wrong place).
- After the cookie is set, the page reloads once so already-broken image requests are retried with the cookie attached.

Fixes #2752

## Test plan
- [x] `npm run lint`
- [x] `test/server-utils.test.js` — URL rewrite, auth-script injection, probe-path extraction
- [x] `test/server.test.js` — `content/ serving` and `da.live login redirect` suites (redirect validation incl. open-redirect rejection)
- [x] `test/content/da-auth.test.js` — `startDaLoginRedirect`
- [ ] Manually verified end-to-end against a real `aem content clone` + `aem up` project (images broken → auth redirect → images load; reload not required after cookie is set)


tripodsan:
I cannot judge if this is correct. looks way too complicated.... why are the image references on content.da.live ?

kptdobe:
> I cannot judge if this is correct. looks way too complicated.... why are the image references on content.da.live ?

This is how DA works today: when you d&d an image inside a "page" document, image is uploaded inside a ".page" folder next to "page" and is referenced via content.da.lve.

And yes, it is really complicated.

tripodsan (dismissed):

Issue #2756, opened 2026-07-06
[content] Local dev (aem up) doesn't process section-metadata blocks for DA content


shsteimer:
**Is your feature request related to a problem? Please describe.**

[aem-lib#213](https://github.com/adobe/aem-lib/pull/213) removed client-side `section-metadata` handling; it's now handled server-side in `helix-html-pipeline` (`extract-section-metadata.js`, [adobe/helix-html-pipeline#1059](https://github.com/adobe/helix-html-pipeline/pull/1059)), gated by `features.rendering.version`. helix-cli has no equivalent, so `aem up` shows a raw, un-decorated `div.section-metadata` block when previewing DA content locally instead of the styled section it renders as in production.

**Describe the solution you'd like**

Add a serve-time transform in `HelixServer.js`: `div.section-metadata` → classes (from `style`) + `data-*` attributes on the parent section, block removed. Apply this unconditionally, without mirroring the `features.rendering.version` flag: not every developer has read access to site config, and the visual result is the same either way (pre-#213 client-side `decorateSections` and the new server step compute equivalent output from the same block). [adobe/da-live#971](https://github.com/adobe/da-live/pull/971) backs this up: DA's own live preview/Quick Edit/Experience Workspace now processes section-metadata in `prose2aem.js` unconditionally too, with no gate.

Prefer reusing [`@adobe/helix-html-pipeline`'s `src/steps/extract-section-metadata.js`](https://github.com/adobe/helix-html-pipeline/blob/main/src/steps/extract-section-metadata.js) directly rather than hand-porting the logic. Not part of the package's public API, so this is an internal import.

**Describe alternatives you've considered**

`da-admin`/pre-push processing: rejected to keep `content/` a lossless mirror of DA storage.
Fetching site config to mirror the flag exactly: rejected, since not every developer has read access to it.

Related: #2755 (same gap for icons).

PR #2759, opened 2026-07-06
feat(content): rewrite :icon-name: syntax in local dev

### LINKED ISSUE #2755: [content] Local dev (aem up) doesn't process :icon-name: syntax for DA content

`aem up` serves DA-cloned content directly from `content/`, which holds raw, unprocessed HTML (the same format DA stores). `helix-html-pipeline`'s `rewrite-icons.js` converts `:icon-name:` text to `<span class="icon icon-name">` server-side for all content, but helix-cli has no local equivalent. So `:icon-name:` shows as literal text when previewing DA content locally, instead of the icon it renders as in production.

Add a serve-time transform in `HelixServer.js`, alongside the existing page-metadata handling (`content-metadata-html.js`): `:icon-name:` → `<span class="icon icon-name">`.

Prefer reusing [`@adobe/helix-html-pipeline`'s `src/steps/rewrite-icons.js`](https://github.com/adobe/helix-html-pipeline/blob/main/src/steps/rewrite-icons.js) directly (small, pure HAST-in/HAST-out function) rather than hand-porting the regex. Not part of the package's public API, so this is an internal import; worth asking html-pipeline to export it officially.

Doing this in `da-admin` or as a pre-`push` processing step: rejected, since `content/` should stay a byte-for-byte mirror of DA storage (confirmed via `da-api.js`, and DA's own `prose2aem.js`, which only applies this transform in its own `livePreview` mode).

Related: #2756 (same gap for `section-metadata` blocks).

shsteimer:
## Summary
- `aem up` serves DA-cloned content directly from `content/`, which holds raw HTML in the same unprocessed form DA stores it. `helix-html-pipeline` rewrites `:icon-name:` text into `<span class="icon icon-name">` server-side for all content, but helix-cli had no local equivalent, so `:icon-name:` showed as literal text when previewing locally.
- Adds a serve-time transform in `HelixServer.js`, alongside the existing page-metadata handling, that mirrors this.
- Reuses `@adobe/helix-html-pipeline`'s own `src/steps/rewrite-icons.js` directly against a locally parsed HAST tree, rather than hand-porting the regex, so local behavior can't drift from what production actually does.

Fixes #2755. Follow-up for `section-metadata` blocks tracked in #2756.

## Test plan
- [x] TDD: added failing unit tests in `test/content-icons-html.test.js` before implementing `src/content/content-icons-html.js`, then made them pass
- [x] TDD: added a failing integration test in `test/server.test.js` before wiring the transform into `HelixServer.js`, then made it pass
- [x] `npm run lint`
- [x] `npm test`

### DIFF
--- a/src/content/content-icons-html.js
+++ b/src/content/content-icons-html.js
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/**
+ * Rewrites da.live-style `:icon-name:` text in content HTML into
+ * `<span class="icon icon-name"></span>` for local dev (`aem up`), mirroring
+ * the same transform helix-html-pipeline applies server-side in production.
+ */
+
+import { unified } from 'unified';
+import rehypeParse from 'rehype-parse';
+import { toHtml } from 'hast-util-to-html';
+// eslint-disable-next-line import/no-unresolved, import/extensions
+import rewriteIcons from '@adobe/helix-html-pipeline/src/steps/rewrite-icons.js';
+
+const REHYPE_PARSE = { fragment: true };
+
+/**
+ * @param {string} htmlFragment body-only or partial HTML from content/
+ * @returns {string} the fragment with `:icon-name:` text rewritten to icon spans
+ */
+export function transformContentIconsHtml(htmlFragment) {
+  let tree;
+  try {
+    tree = unified().use(rehypeParse, REHYPE_PARSE).parse(htmlFragment);
+  } catch {
+    return htmlFragment;
+  }
+
+  const before = toHtml(tree);
+  rewriteIcons({ content: { hast: tree } });
+  const after = toHtml(tree);
+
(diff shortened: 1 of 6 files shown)


src/content/content-icons-html.js
```diff
@@ -0,0 +1,45 @@
...
+ * Rewrites da.live-style `:icon-name:` text in content HTML into
+ * `<span class="icon icon-name"></span>` for local dev (`aem up`), mirroring
+ * the same transform helix-html-pipeline applies server-side in production.
+ */
+
+import { unified } from 'unified';
+import rehypeParse from 'rehype-parse';
+import { toHtml } from 'hast-util-to-html';
+// eslint-disable-next-line import/no-unresolved, import/extensions
+import rewriteIcons from '@adobe/helix-html-pipeline/src/steps/rewrite-icons.js';
```
shsteimer, line 23: flagging this deep import for review/discussion. Would it make sense to change `helix-html-pipeline` to provide these as exports for re-use? Note we'll need a similar thing to handle #2756

tripodsan:
I think this gets more and more complicated. next, you need metadata injection and the section metadata decoration etc... maybe it would be better to run the `html2md -> html-pipeline` chain instead.

however, is this really a usecase to render from a DA content checkout?

shsteimer:
@tripodsan 

> however, is this really a usecase to render from a DA content checkout?

exactly. DA checkout content is stored in format author will use, in this case `:icon-name:`. but html for preview needs something slightly different `<span class="icon icon-name>...</span>`

It looks like we already do this for page metadata blocks (see https://github.com/adobe/helix-cli/blob/main/src/content/content-metadata-html.js), but the same issue also exists for section metadata blocks since those are now transformed in the pipeline (https://github.com/adobe/helix-cli/issues/2756). So agreed this starts to get complicated. But I'm not sure where else to handle these differences between html-pipeline markup and authored content.

leetsao:
I've run across this in local environments where key elements are injected via :logo: syntax.  This makes local editing hard to use because icons are not rendered locally and are off key to site view.

tripodsan:
> But I'm not sure where else to handle these differences between html-pipeline markup and authored content.

as I said, we should use the entire html2md->html-pipeline chain to render the DA markup.

PR #2761, opened 2026-07-09
feat(content): render DA content via the real html2md -> html-pipeline chain

### LINKED ISSUE #2755: [content] Local dev (aem up) doesn't process :icon-name: syntax for DA content

### LINKED ISSUE #2756: [content] Local dev (aem up) doesn't process section-metadata blocks for DA content

shsteimer:
* When serving local DA checkout content in the `content/` directory, run those through the real pipeline: `content/*.html` → `@adobe/helix-html2md` → Markdown → `@adobe/helix-html-pipeline`
* This handles icons, section metadata, page metadata all through the same pipeline used when publishing content, instead of selectively re-implementing the pieces we need (e.g. the now deleted `src/content/html-pipeline-internals.js`

- [x] when merging / squashing, make sure the fixed issue references are visible in the commits, for easy compilation of release notes

## Related Issues

Fix #2755 
Fix #2756 
Close #2759 (superseded by this)

Thanks for contributing!


tripodsan:
cool. but this looks rather complicated, why not directly use the htmlPipe like here, for example:
https://github.com/adobe/helix-html2md-lib/blob/main/test/roundtrip/mock-pipeline.js

shsteimer:
> cool. but this looks rather complicated, why not directly use the htmlPipe like here, for example: https://github.com/adobe/helix-html2md-lib/blob/main/test/roundtrip/mock-pipeline.js

Thanks. as I was doing, I figured there'd be a simpler way; Updated with this approach.

tripodsan (approved):

